PR #21, opened 2020-03-14
Add a Fourth Cause to SCEP NDES Troubleshooting

nbuuck:
Add an additional and likely-common cause of SCEP app pool crashes, as the certificate referenced by the NDESPolicy module is bound to expire but present in a non-obvious way to administrators (w3wp crash referenced ntdll.dll as the faulting module). This was sourced from a real case I resolved after an NDES server had been in service for one year.

@brenduns (doc author)


PRMerger8:
@nbuuck : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Brenduns:
#sign-off

PR #40, opened 2020-03-19
Update device-enrollment-program-enroll-ios.md

AnyaNovicheva1:
We should add a note based off of new documentation that Jack Poehlman brought up. 
Based on this documentation from Apple - https://support.apple.com/guide/apple-business-manager/edit-mdm-servers-asm54ab8fff2/web

A user with the proper privileges may sometimes need to edit MDM server information. For example, a user must replace the active server token on an MDM server in these situations:
•	When a new public key is created or if a new server token is generated
•	When the Managed Apple ID password of the user who downloaded the initial token is changed
•	As a security measure, when the user who downloaded the original token leaves your organization

### DIFF
--- a/memdocs/intune/enrollment/device-enrollment-program-enroll-ios.md
+++ b/memdocs/intune/enrollment/device-enrollment-program-enroll-ios.md
@@ -285,7 +285,8 @@ See [Enroll your iOS/iPadOS device in Intune with the Device Enrollment Program]
 
 ## Renew a DEP token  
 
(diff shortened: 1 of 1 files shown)


PRMerger15:
@AnyaNovicheva1 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger15:
@dougeby : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #46, opened 2020-03-20
Need more prescriptive guidance about Defender AV exclusions

jeffgilb:
It's not necessary to add exclusions for Defender AV for greenfield management of Windows 10 with Intune, but we don't say that anywhere like we do for the Server OS's. Adding this content (from those other docs) to the Intune content to make it more explicit.


PRMerger9:
@jeffgilb : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #195, opened 2020-05-05
Update certificates-profile-scep.md

anpais:
Some settings are required from apple as we can see on Apple site.


PRMerger17:
@anpais : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Brenduns:
@anpais   Thanks for the submission!  Before I can accept it, we need a few edits, which I can't introduce to your PR directly.  

•	iOS and macOS should appear with the first letter as lowercase.  
•	Also, we need to remove the “en-us” bit from the URL when possible. The URL works fine without that. While its not clear that Apple has localized versions of this article, if they do, removal of a language specific path ensures customers get content that matches their browsers default language version, when content is available in that language version. 
•	Finally, a bit of rewording to imply the link opens a non-Intune or Microsoft page.  

When done, that new bullet item should appear as follows:

- On iOS 13 and macOS 10.15, there are some [additional security requirements that are documented by Apple](https://support.apple.com/HT210176) to take into consideration.

Brenduns:
@anpais  Never mind. I see this was closed while I had it open, and a new PR arrived. I'll respond in that.

PR #308, opened 2020-05-27
Fixed formatting of command examples

AdamGrossTX:
The commands had - in them that aren't needed and also were all on a single line.


PRMerger13:
@AdamGrossTX : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Brenduns:
#sign-off   Looks like these were really missing spaces between the dash and the first * for each entry, but trailing spaces work as well.

PR #332, opened 2020-06-04
.NET Framework 3.5 is a pre-requisite for the portals

PanuSaukko:
You should add .NET Framework 3.5 as a pre-requisite for the MBAM portals. 

MBAMWebSiteInstaller.ps1 will try to install the required Windows server roles and features. One of the required features is WAS-NET-Environment, which requires .NET Framework 3.5. Because .NET Framework 3.5 installation requires SXS\microsoft-windows-netfx3-ondemand-package....cab file from the OS media, PowerShell script cannot install .NET Framework 3.5  And PowerShell will give you a cryptic error message that the installation fails. 

Other roles and features can be installed without files from OS media and thus those don't need to be listed as a pre-req.


PRMerger17:

aczechowski:
#sign-off

PR #333, opened 2020-06-04
Added more clearly that AD joined device is required

PanuSaukko:
You say that: "on-premises Windows clients" are required. I'd like a more precise statement and the client needs to be joined to Active Directory and AAD joined/workgroup clients are not supported.


PRMerger7:

aczechowski:
#sign-off 

*Once merged*, this should be live with the next daily publishing process. Right now that *should* be tomorrow (Friday, 5 June). No further action needed from you.

PR #340, opened 2020-06-07
Updated language to reflect client settings policy

AdamGrossTX:
This is the original client setting that was referred to.
image

Here's what it looks like in 2002 (and previous versions). I don't know when it actually changed, but based on the 1902 note for the other verbiage, I matched it.

image


PRMerger17:
@AdamGrossTX : Thanks for your contribution! The author(s) have been notified to review your proposed change.

mestew:
@AdamGrossTX Thanks Adam! Looks like I forgot to update this file when I made that 1902 change =) 

#sign-off

PR #353, opened 2020-06-10
Dedupe folder vs volume support

ChadSimmons:
Per https://web.archive.org/web/20140310081022/http://blogs.technet.com/b/configmgrteam/archive/2014/02/18/configuration-manager-distribution-points-and-windows-server-2012-data-deduplication.aspx and other sources, the **FOLDER**(s) that host package source files can't be marked for data deduplication, not the **VOLUME**(s).

### DIFF
--- a/memdocs/configmgr/core/plan-design/configs/support-for-windows-features-and-networks.md
+++ b/memdocs/configmgr/core/plan-design/configs/support-for-windows-features-and-networks.md
@@ -64,7 +64,7 @@ Configuration Manager provides support for clients in workgroups.
 Configuration Manager supports the use of data deduplication with distribution points on Windows Server 2012 or later.
 
 > [!IMPORTANT]  
-> The volume that hosts package source files can't be marked for data deduplication. This limitation is because data deduplication uses reparse points. Configuration Manager doesn't support using a content source location with files stored on reparse points.  
+> The folder(s) that host package source files can't be marked for data deduplication. This limitation is because data deduplication uses reparse points. Configuration Manager doesn't support using a content source location with files stored on reparse points.  
 
 For more information, see the following posts:
 


baardhermansen:
@ChadSimmons According to the same article, there's also a note regarding volume:

> Note: Because Data Deduplication internally uses reparse points, and Configuration Manager does not support using a content source folder with files stored on reparse points, therefore using a content source folder **located on a volume enabled for Data Deduplication may not work**.

ChadSimmons:
@baardhermansen, thank you for the reply.

The vagueness of the note is the issue.  Johan Arwidmark even commented on the original blog post about this point and it was not answered there either.  The crux of the detail is what folders can/cannot be enabled for deduplication.  I suppose the easy way out is to simply state that it is not supported to enable deduplication on a volume where content source files exist; however, to support the broadest range of scenarios a detailed folder list would be helpful.  Without better guidance the community will have to lean on MVPs [Johan](https://deploymentresearch.com/using-data-deduplication-with-configmgr-2012-r2), [Rob Marshall](https://archive.wmug.co.uk/wmug/b/r0b/posts/windows-2012-server-deduplication-and-configmgr-2012), and others who state that only the **SCCMContentLib** and **SMSPKGx$** folders can be deduplicated and *everything* else should be excluded.

baardhermansen:
I'm fairly certain i've seen a blog post about which folders are available for dedup. Provided the source files resides on a a different volume, there should be no issues activating dedup.

baardhermansen:
Found the blog: https://techcommunity.microsoft.com/t5/configuration-manager-archive/configuration-manager-distribution-points-and-windows-server/ba-p/273385

So it seems then that dedup can be activated on the DPs, without folder exceptions.

ChadSimmons:
I was hoping Microsoft would clarify the technical details in the official documentation rather than leaving it vague with the "may not work" statement and the incorrect statement that "The volume that hosts package source files can't be marked for data deduplication".  As noted by the MVP guidance, a volume that hosts both the Content Library and Content Source can be deduplicated as long as the Content Source doesn't contain reparse points.  The MVP guidance will just have to be good enough.

PR #354, opened 2020-06-10
Corrected terms

allanhirt:
You were using incorrect terms here. SQL Server cluster is not a thing, etc. I will also be changing the AG ones as well to reflect the right terms.

Also, you should never just say SQL. It's SQL Server. Even SQL Server sometimes get this wrong, but marketing does not like it. 

You have links to SQL Server doc that needs to be corrected as well, and I just submitted all of those. I think I just did 4 or 5 PRs.

It's a Windows Server failover cluster (WSFC), not WIndows failover cluster. Failover clustering is the name of the underlying feature.

You cannot say virtual instance. That was basically banned in 2008 when Virtual Server was named Hyper-V and virtual servers became a "thing".
Saying "clustered SQL Server instance" is OK, though.

### DIFF
--- a/memdocs/configmgr/core/servers/deploy/configure/use-a-sql-server-cluster-for-the-site-database.md
+++ b/memdocs/configmgr/core/servers/deploy/configure/use-a-sql-server-cluster-for-the-site-database.md
@@ -1,7 +1,7 @@
 ---
-title: SQL Server cluster
+title: Always On Failover Cluster Insatnce
 titleSuffix: Configuration Manager
-description: Use a SQL Server cluster to host the Configuration Manager site database
+description: Use an Always On failover cluster instance to host the Configuration Manager site database
 ms.date: 04/30/2020
 ms.prod: configuration-manager
 ms.technology: configmgr-core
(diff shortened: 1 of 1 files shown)


aczechowski:
Thanks @allanhirt for the feedback. I need to do some checking on this. We have separate articles for SQL Server Always On. This article is older and generally untouched for a while, so it's possible that the technology has evolved past what this describes.

allanhirt:
Always On is not a feature, so if you have an article on that it’s wrong. The feature is availability groups. The full formal feature name is Always On Availability Groups. Always On is an umbrella with TWO features: AGs and FCIs.

Sounds like I may have more editing if you’re referring to AGs as Always On which is wholly incorrect.

aczechowski:
I got some input from engineering, but I want to sync with Masha on the SQL team (to whom you filed several PRs as well) to make sure we're aligned.

I want to let you know I haven't forgotten about this! 😄

aczechowski:
Hi @allanhirt, I finally got back to this after getting input from our engineering team and SQL docs team. I pulled your commit over to our private repo, and then made changes across our entire doc library (139 files) for consistent and accurate SQL branding. Your contribution (commit) is still included. I hope I'll finalize and get it through review tomorrow. Thanks again for your contribution! I'm closing this PR as it's no longer needed.

PR #404, opened 2020-06-22
Specify key size for media

frankroj:
Adding key size used when encrypting environment variables in media. Check in code and key size is 128


PRMerger13:
@frankroj : Thanks for your contribution! The author(s) have been notified to review your proposed change.

aczechowski:
#sign-off

PR #670, opened 2020-08-24
Update whats-new.md

ShrCaJesmo:
Adds a note to the entry "Admins no longer require an Intune license to access Microsoft Endpoint Manager admin console" specifying a license may still be required for some functions. After testing, it seems an Intune License is still required to add the Teamviewer Connector flow. I've been told this limitation is by design, in which case it would be best to document some of these features may still not work by design.


PRMerger12:
@ShrCaJesmo : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ErikjeMS:
#sign-off

PR #696, opened 2020-08-28
Clarifying purpose of OEMConfig email alias

msft-jy:
We keep getting emails from people who don't represent OEMs, trying to clarify that the alias is not a customer support alias.


PRMerger14:
@msft-jy : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #741, opened 2020-09-09
Updated text for clarity

AdamGrossTX:
The wording wasn't clear which user was being referred to. Obviously this is a preview feature, so I'm sure the docs will be revised before GA, but figured this would help those wanting to onboard now.


PRMerger17:
@mestew : Thanks for your contribution! The author(s) have been notified to review your proposed change.

mestew (approved):

mestew:
@AdamGrossTX Thanks so much for the contribution. It added some much needed clarification. I updated the troubleshooting articles too using your suggestion. I also giggled when I noticed that you have the same exact problem I do...I always want to say "admin console" rather than "admin center" for MEM. Old habits totally die hard ;) 

#sign-off

PR #754, opened 2020-09-11
Updated to remove duplicate info

rudybankson:
Updated to remove duplicate information on security software blocking Run Scripts and CMPivot.


PRMerger17:
@rudybankson : Thanks for your contribution! The author(s) have been notified to review your proposed change.

mestew:
@rudybankson  Hey Rudy, Normally I'd be all about removing a duplicate. I dug back through the history of why there are 2 bullets that are so similar. This was actually done on purpose for a very specific reason. We needed to distinguish between AV/malware scanners v/s security systems that are doing things like detecting potential lateral movement paths. Basically, it's to note that both of these should be accounted for by CM and security admins so they could make proper decisions for their org. 

#please-close

PR #757, opened 2020-09-11
Update data-transfer-between-apps-manage-ios.md

msintuneappsdk:
Currently, there is no support for enrolling with a different user on an app if there is a MDM enrolled account on the same device.


Erikre:
#sign-off

PR #795, opened 2020-09-23
Update security-baseline-settings-mdm-all.md

maurice-daly:
Windows Hello for Business settings are no longer available in the September security baseline

### DIFF
--- a/memdocs/intune/protect/security-baseline-settings-mdm-all.md
+++ b/memdocs/intune/protect/security-baseline-settings-mdm-all.md
@@ -2245,65 +2245,6 @@ For more information, see [Policy CSP - WindowsConnectionManager](/windows/clien
 ::: zone-end
 ::: zone pivot="mdm-may-2019,mdm-sept-2020"
 
-## Windows Hello for Business
-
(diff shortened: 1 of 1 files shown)


PRMerger6:
@maurice-daly : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Brenduns:
@maurice-daly  Thanks for this contribution. I regret we'll not be able to accept this, as this new baseline has a few more complex issues which required other edits. Those will all publish later today. 

In addition to _Windows Hello_ having been removed, _Above Lock_ was also remove.  Also, for awareness, the baseline files use a system of zone-pivots, which are settings that declare which baseline version a portion of the file (settings) appears in. These can be singular versions, some, or all versions.  Your edit did remove the Windows Hello settings, but would have removed them from all baseline versions.  Instead, due to how this file is constructed, a more surgical edit is needed to edit the pivot statements to remove the Windows Hello (and Above Lock) from the 2020 August version.  

I hope that makes sense, but again, I very much appreciate anyone who takes it on themselves to contribute through edits. This file however held a bit more unexpected churn than just the one section.

PR #802, opened 2020-09-23
Indent caused image not to display

wattbt:
Removed extra indent on connect-wsus-console.png image which was I think causing it not to display in the article, and instead appearing as a code block.


PRMerger19:

mestew:
Thanks for noticing! I included this fix when we added links to the community hub PowerShell scripts in this article. 
#please-close

PR #807, opened 2020-09-24
Updated the app assignment flow to call out Device Licensing

rogersoMS:
As this is an offline app, the licensing needs to be changed from the default (User) to device licensing. Although we provide a link to the page on how to assign apps, I thought it would be good to provide a couple of more granular steps here so the user can keep in the flow. I'd like to bold some text, but I cannot  :-(

Bold text: Properties, Assignments, Edit, Add Group Device licensing 'Review + save'


PRMerger14:
@rogersoMS : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ktoliver:
@rogersoMS Hi. This pull request was opened in the public repo. PMs should work in the private repo, per the Microsoft Docs contributor guide. Would you make this content update in the private repo? We'll close this PR. Thanks.

PR #811, opened 2020-09-26
Update lob-apps-windows.md

kkaminsk:
Link points to the old and no longer available old Intune console.


PRMerger6:
@kkaminsk : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Erikre:
@kkaminsk - Thank you for the suggestion. We have one main link in the Intune docs when signing in to Microsoft Endpoint Manager. It should take you directly to the Microsoft Endpoint Manager admin center. It should no longer take you to the old portal. I checked this link and it seems to work fine.

PR #848, opened 2020-10-07
Fixing redirection for windows-enrollment-company-portal

kagillig:
The Windows Company Portal used to use the https://docs.microsoft.com/en-us/mem/intune/user-help/windows-enrollment-company-portal link and we noticed it now leads to a 404 page. When reviewing this file, it looks like the source_path here is missing the -portal, unless there was previously another link that was just windows-enrollment-company. If so, this is the wrong change and I'll just need to add a separate redirection item. Let me know if that is the case.

### DIFF
--- a/.openpublishing.redirection.json
+++ b/.openpublishing.redirection.json
@@ -16,7 +16,7 @@
             "redirect_document_id": false
         },
         {
-            "source_path": "memdocs/intune/user-help/windows-enrollment-company.md",  
+            "source_path": "memdocs/intune/user-help/windows-enrollment-company-portal.md",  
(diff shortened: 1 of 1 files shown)


lenewsad:
#sign-off

PRMerger14:
@lenewsad: I'm sorry - only the author of this article, @, can sign off on your changes. But we do have an exception process - if you are on the Microsoft content or product team for this product area, you can ask the PR review team to review and merge it by sending mail to the techdocprs alias.

lenewsad:
@kagillig, I opened a PR in the MSFT employee repo to resolve the above error message. Closing this one.

PR #869, opened 2020-10-13
Update samsung-knox-apps-allow-block.md

Othou:


PRMerger6:
@Othou : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MandiOhlinger:
@Othou Thank you! Why do you think it should be "block" instead of "allow"? Just trying to understand. Did your testing prove this? Any additional information would be great.

Othou:
Hey @MandiOhlinger! I have not tested this, but I would assume it should say block because it is example config for: "For a list of apps that are blocked from running on the device", as row 52 says.
Also on row 56 "..PreventStartPackages..", it indicates that it is blocking an application package.
That's why I think row 58 should say block, since we are doing "a list of apps that are blocked"

MandiOhlinger (approved):

MandiOhlinger:
@Othou You are correct. Thank you! #sign-off

PR #950, opened 2020-10-30
CMG connection point --> SUP ports are incorrect

PanuSaukko:
You need to open 8530/8531 or 80/443 port from CMG connection point to SUP depending on the ports WSUS is using.  Currently 8530/8531 ports are missing.


PRMerger14:

PR #1024, opened 2020-11-18
Make hash extraction script more PowerShell-y

KennethStebbins:
Instead of using "md" to create a directory (and using an escaped backslash in the path), the script uses New-Item and encapsulates the path in quotes.
The path used with Set-Location is now encapsulated in quotes.


PRMerger20:
@KennethStebbins : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #1030, opened 2020-11-19
Add Automatic Client Upgrade

bryandam:
This has come up a couple of times now given that our WFH situations have dragged on this long.  Users are concerned about updating clients over the CMG.  If you squint real hard you'll realize that ACU _is_ Software Distribution but ... sigh ... let's make this super obvious for people.

No idea how you have this list sorted so feel free to move it wherever you want in the list.

### DIFF
--- a/memdocs/configmgr/core/clients/manage/cmg/supported-configurations.md
+++ b/memdocs/configmgr/core/clients/manage/cmg/supported-configurations.md
@@ -36,7 +36,6 @@ The following table lists CMG support for Configuration Manager features:
 
 | Feature | Support |
 |---------|---------|
-| Automatic Client Upgrade | ![Supported](media/green_check.png) |
 | Software updates | ![Supported](media/green_check.png) |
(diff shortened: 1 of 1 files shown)


PRMerger18:

PRMerger8:
@aczechowski : Thanks for your contribution! The author(s) have been notified to review your proposed change.

aczechowski (approved):
Thanks @bryandam for the contribution! Agreed, good to be explicit. I moved it down the list next to the client install rows.

aczechowski:
#sign-off

ktoliver:
Merging the PR.

@aczechowski The "not supported" Xs don't show up in GitHub Preview because of a difference in casing in the file name between the MD image reference and the file name in the repo. It's not critical, but it would be better to have the images display than not. Maybe something for a future update? Thanks.

image

aczechowski:
Thanks @ktoliver, we have a bug to revisit all of our icons because they don't show well in dark mode. so I'll add this note to the list. 😄

PR #1037, opened 2020-11-20
Azure portal is no longer used for Intune

reiikei:
Azure portal is no longer used for Intune


PRMerger14:
@reiikei : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #1055, opened 2020-11-23
Update Autopilot Redeployment/Reset Setting

angela-em:
The setting has been renamed within the Intune Device Restriction policy to "Autopilot Reset".  When customers are enabling this setting, they may not understand that it is actually enabling the local Autopilot reset option.


PRMerger17:
@angela-em : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #1084, opened 2020-12-03
Update compliance-policy-create-windows.md

SaiDuanZX:
We have test and confirmed when we try to check with EncryptionCompliance status from Windows 10 devices, we will only check with  Win32_EncryptableVolume , and this WMI only support with BitLocker, as you may check in https://docs.microsoft.com/en-us/windows/win32/secprov/win32-encryptablevolume , so it means our "Encryption of data storage on a device" check in Compliance may only support with BitLocker Encryption, if we do not state this clearly, customer may be confused and request us why the 3rd party encryption status cannot be detected


PRMerger18:
@SaiDuanZX : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger6:
@Brenduns : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Brenduns:
#sign-off

PR #1163, opened 2020-12-23
Adding note about AAD CA policies and excluding "Microsoft Intune"

JeremyTBradshaw:
I copied / pasted the block note directly from the [Set up Intune enrollment of Android Enterprise fully managed devices](https://docs.microsoft.com/en-us/mem/intune/enrollment/android-fully-managed-enroll) page.  I just verified in testing, it 100% is still necessary and so that blurb needs to be on this page.

For what it's worth, at one point in time, we would only need to exclude the "Microsoft Intune Enrollment" cloud app from the CA policies that are requiring compliant device or blocking.  Nowadays, it is 100% definitely the "Microsoft Intune" cloud app that needs to be excluded, and excluding the "Microsoft Intune Enrollment" cloud app has no effect at all.  I wish the people making these changes in the product would take the time to update the documentation when they make the changes.

### DIFF
--- a/memdocs/intune/enrollment/android-dedicated-devices-fully-managed-enroll.md
+++ b/memdocs/intune/enrollment/android-dedicated-devices-fully-managed-enroll.md
@@ -44,6 +44,9 @@ After you've set up your Android Enterprise [dedicated devices](android-kiosk-en
 > [!TIP]
 > Corporate-owned work profile (COPE) device management is available on Android version 8.0 and newer.
 
+> [!NOTE]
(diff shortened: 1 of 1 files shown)


PRMerger7:
@JeremyTBradshaw : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ErikjeMS:
#sign-off

jborsecnik:
@ErikjeMS, the added link doesn't work in preview. Are you sure it will work when this is published? Thanks.

ErikjeMS:
@JeremyTBradshaw Can you make sure that the link will work in production?

ErikjeMS:
@jborsecnik Yes, it will work. It currently works in the following published article: https://docs.microsoft.com/en-us/mem/intune/enrollment/android-fully-managed-enroll

JeremyTBradshaw:
Thanks @ErikjeMS , I couldn't get to it earlier.

PR #1170, opened 2020-12-28
Consistent platform properties section

illumiN8i:
As a new user to Intune following this documentation, I didn't realize that I needed to select Android Personally-Owned Work Profile specifically when creating the trusted certificate profile leading to me opening a support case with Microsoft. This change copies the text from the PKCS section further down the page to make it consistent within the document.


PRMerger16:
@illumiN8i : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Brenduns:
#sign-off       @illumiN8i  Thanks for this submission.

PR #1197, opened 2021-01-08
Need to manually open port 443 on firewall

PanuSaukko:
You need to manually open HTTPS port 443 on your firewall if you are using AdminService without IIS in version 2010.


PRMerger19:

aczechowski (approved):
Thanks @PanuSaukko for the contribution, good point to include. I don't think it's in the ports article already, so I'll get to that too.

aczechowski:
#sign-off

PR #1198, opened 2021-01-08
Both site servers need Full control permissions to System - System Ma…

PanuSaukko:
…nagement container

Both site servers need Full control permissions to Active Directory's System - System Management container. The 1st site server already has it, but the permissions need to be added to the passive site server.


PRMerger15:

aczechowski (approved):
Thanks @PanuSaukko for the contribution! I'll add a note as well to the article that talks about permissions on that container.

aczechowski:
#sign-off

PR #1225, opened 2021-01-14
Update ios-user-enrollment-supported-actions.md

ShrCaJesmo:
I've been instructed to document the issue here. Results of multiple attempts at assigning VPP apps. Each time, the ones assigned during the previous user enrollment would work in future enrollments after passing by device enrollment, but newly targeted apps on user-enrolled device never stop throwing error. As far as I can tell, Company Portal can't pull the configured AppleID across the sandbox barrier. Never was able to get the data I needed to validate this is the issue but I suspect Intune isn't providing associateClientUserIdStrs property when doing manageVPPLicensesByAdamIdSrv per https://developer.apple.com/business/documentation/MDM-Protocol-Reference.pdf pg159. It'd be nice if we could see the result of the Intune<>Vpp traffic somewhere, or maybe I just don't know where to look.


PRMerger6:
@ShrCaJesmo : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ErikjeMS:
#sign-off

PR #1278, opened 2021-01-30
Change Only when no user is logged on

Eri-Ichikawa:
This is wrong. We can enable "Allow users ..." when  select "Only when a user is logged on" or "Whether or not a user is logged on" on Create Package and Program Wizard.


PRMerger13:
@Eri-Ichikawa : Thanks for your contribution! The author(s) have been notified to review your proposed change.

aczechowski (approved):
Good catch, thanks @Eri-Ichikawa. It's probably been like that for years. 😟

aczechowski:
#sign-off

PR #1288, opened 2021-02-02
Update pki-certificate-requirements.md

REPACH:
Key Usage value must contain Digital Signature, Key Encipherment (a0).
Added to the PKI certificates for clients.

Key Encipherment is needed on the Key Usage in order to use BITS Helper-Tokens (https://docs.microsoft.com/en-us/windows/win32/bits/helper-tokens-for-bits-transfer-jobs) In windows 10, because BITS uses the certificate to encrypt/decrypt test data to verify the user has access to the certificate.


mestew:
#sign-off

mestew (approved):

mestew:
@REPACH Thanks for the doc change! Your first PR against this repo it totally approved and is perfect.

PR #1305, opened 2021-02-08
Add extra known issue for GPO enrollment in ESP scenario

leonzhuyihong:
According to https://portal.microsofticm.com/imp/v3/incidents/details/214735469/home 
"What is happening here (and what Britini was referring to earlier) is that the method by which the customer is enrolling devices (Group Policy) will never allow ESP to run, meaning it will never be marked completed. However, because the customer wants ESP to be enabled regardless, the policy enabling ESP reaches the machine and stays there even though ESP will not ever run.
When the Sidecar agent runs, it looks at the ESP policy and sees that it is enabled (even though ESP has not ever and will never run on the machine) so Sidecar will wait indefinitely for ESP to be marked completed in order to provision any apps not on the ESP blocking list that is configured in the ESP profile. However, ESP will never be marked completed because it will never run, thus the apps are never provisioned. This is all by design."
Suggest to add this into doc for other customers can notice it


PRMerger18:
@leonzhuyihong : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ErikjeMS:
Investigating with PM...

ErikjeMS:
#sign-off

PRMerger20:
@ktoliver : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ktoliver:
@ErikjeMS Is my grammar edit correct technically?

ErikjeMS:
@ktoliver Looks good except you could use "with" instead of "via".

PRMerger14:
@ktoliver : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ktoliver:
Thanks @ErikjeMS; merging.

PR #1422, opened 2021-03-18
Adding WAV as supported Audio file

brianreidc7:
As per this "Completed" comment in User Voice (https://microsoftintune.uservoice.com/forums/291681-ideas/suggestions/12403098-support-playing-wav-files-in-outlook) the AIP App now does WAV files and this works seemlessly when listening to voicemails in emails on MAM protected Android devices.


PRMerger17:
@brianreidc7 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

lenewsad:
Hi @brianreidc7, thanks for your contribution! I double-checked with the AIP app team and WAV files are not supported so we shouldn't add this information to the docs.

PR #1433, opened 2021-03-22
Help me Win Online Arguments about Client Cache

bryandam:
... cause let's be honest.  That's why we edit docs ... is it not?

It's long been my understanding that there is no mechanism within ConfigMgr that removes items from the client cache 'just because'.  You set the cache size, the client fills the cache to that size as it download content, and from that point on only removes items if it needs room in the cache to download new content.  Which is to say: that within reason the cache will hover around the configured max value over time. Fluctuating because the items removed might be larger that the new items coming in.  So many orgs have implemented their own 'maintenance' solutions to this 'problem' that people have started just assuming that's how the product works instead of something someone put in their environment years ago.

This doc _kind_ of hinted at that but not clearly enough to earn me street cred on Reddit.  Wait ...  I mean ... to properly aid users in understanding and using the product.  Yea, that's it.

Though, lest you think I'm just in it for the 'gram: I updated the doc to reflect the functionality added in 1906 (now out of support) that made the cache tombstone value configurable.  It's no longer hard-coded at 24 hours like this doc originally stated.

I suck at the English language (as well as every other language humans have invented) so please feel free to wordsmith as you desire.  As long as I still win.  Crap, I did it again. I mean, as long as clarity and great justice is brought to mankind.

### DIFF
--- a/memdocs/configmgr/core/clients/manage/manage-clients.md
+++ b/memdocs/configmgr/core/clients/manage/manage-clients.md
@@ -259,7 +259,7 @@ If the client attempts to download content that's less than the size of the cach
 
 If you later increase the cache size, the client attempts to download the content again during the next retry interval. The client tries to download the content every four hours until it tries 18 times.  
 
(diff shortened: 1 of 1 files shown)


PRMerger8:

aczechowski:
Thanks @bryandam, I too think this is correct, but I want to confirm with an expert. Stay tuned...

aczechowski (approved):

aczechowski:
#sign-off

PR #1459, opened 2021-03-29
Update device-restrictions-ios.md

jaye-ren:
Updated for clarification - in MEM, the 'Numeric' option will still allow users to use alphabetic characters, it just doesn't require them to use alphabetic characters. There currently isn't an Apple setting to require numeric only passwords.


PRMerger9:
@jaye-ren : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #1470, opened 2021-04-04
Update pre-provision.md

v-rechau:
Adding a note about disabling MSA preventing the User Flow from happening.

This information is already provided in the Device Restriction page: https://docs.microsoft.com/en-us/mem/intune/configuration/device-restrictions-windows-10#cloud-and-storage

If the behavior of the user flow isn't working as intended, there is a much higher chance admins will try to find out why in this page (or the Known Issue page) rather than searching through device restrictions (or even thinking about them being the cause)


PRMerger6:
@v-rechau : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #1478, opened 2021-04-05
Update certificates-scep-configure.md

davidpaMSFT:
I believe the last two items in the list "To configure a value that can be set from within the Intune console" should be in a separate list/section.


PRMerger13:
@davidpaMSFT : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Brenduns:
#hold-off  Investigating the changes this produces, vs the intent of the original procedure structure.

Brenduns:
@davidpaMSFT   Thanks David, I'm not sure how or when those two procedure got combined, but its good to get them separated out. Much appreciated!

#sign-off

PR #1517, opened 2021-04-13
Update plan-for-software-updates.md

rashid2982:
Added note for Shared DB as we see lot of cases where customers are confused as to why we are picking servers at random

This is because the WSYNCMgr uses PickRandom() from Shared SUP list. There is no way to say which server to go sync with MU, So we advise and recommend customer keep top site independent and share DBs amongst downstreams if they want to.

Added section is this

> [!IMPORTANT]  
>  **Note about using WSUS Shared database scenario:**  
If you are sharing the database of WSUS server on the top site which is talking to Microsoft for updates with other WSUS servers, You will need to make sure all the [internet requirements](https://docs.microsoft.com/en-us/mem/configmgr/sum/plan-design/plan-for-software-updates#BKMK_ConfigureFirewalls) are met for every single WSUS server you are sharing the database with. This is because during sync ConfigMgr randomly chooses the WSUS server amongst the shared ones to sync against Microsoft updates. Our recommendation is to share database amongst the downstream servers and keep the top level servers use its own database for simplicity and avoiding opening more ports and firewall rules than you have to.


PRMerger18:
@rashid2982 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

mestew (approved):
Moved this to another section and did a few minor edits.

mestew:
#sign-off

PR #1533, opened 2021-04-19
Fixing the requirement of JSON data in the table

pamcdo:
I did submit a change that fixed the placement of the JSON data for Android enterprise but left the table set to 'yes' for required data on the device admin and 'no' on android enterprise. I swapped them as well so that it should be correct now.


PRMerger15:
@pamcdo : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ErikjeMS:
#sign-off

PR #1585, opened 2021-05-04
Update on CMG Connection Point

pamcdo:
Previous wording for this section didn't have clear actions for the administrator to execute. I think this update will better enable those reading it to get this piece up and running.

### DIFF
--- a/memdocs/configmgr/core/clients/manage/cmg/setup-cloud-management-gateway.md
+++ b/memdocs/configmgr/core/clients/manage/cmg/setup-cloud-management-gateway.md
@@ -122,7 +122,13 @@ If you're using [client authentication certificates](configure-authentication.md
 
 ## Add the CMG connection point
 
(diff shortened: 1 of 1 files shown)


PRMerger17:
@pamcdo : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger7:
@aczechowski : Thanks for your contribution! The author(s) have been notified to review your proposed change.

aczechowski (approved):
Thanks @pamcdo for the contribution! It's a fair point. We generally try to not duplicate content when possible, but at times it makes sense to do so. Especially in the case of such a popular feature. 😄 

I made some tweaks to the wording for style. For example, we try to avoid talking about "click" or "right-click" as that implies the use of a mouse. We try to be inclusive for all users.

aczechowski:
#sign-off

PR #1593, opened 2021-05-06
Update site-server-high-availability.md

davidpaMSFT:
Add prerequisite for site server in passive mode to have local administrator permissions on all site system servers. This is as per docs here:  https://docs.microsoft.com/en-us/mem/configmgr/core/plan-design/hierarchy/accounts#elevated-permissions, as well observations in live environments (post-installation task has a warning to check logs for missing permissions; as well if site server doesn't have permissions, after it is promoted to active, errors are reported, e.g. Site System Status Summarizer can't access remote site systems).

### DIFF
--- a/memdocs/configmgr/core/servers/deploy/configure/site-server-high-availability.md
+++ b/memdocs/configmgr/core/servers/deploy/configure/site-server-high-availability.md
@@ -97,6 +97,8 @@ Microsoft Core Services Engineering and Operations used this feature to migrate 
 
   - Must have its computer account in the local Administrators group on the site server in active mode.<!--516036-->
 
+  - Must have local administrator rights on all site system servers.  
+
   - Must install using source files that match the version of the site server in active mode.  
 
   - Can't have a site system role from any site installed on it before you install the site server in passive mode role.  


PRMerger12:
@davidpaMSFT : Thanks for your contribution! The author(s) have been notified to review your proposed change.

aczechowski:
Thanks @davidpaMSFT for the contribution suggestion. I'd rather keep those sort of specific requirements in the more general article, [Prerequisites for installing sites](https://docs.microsoft.com/mem/configmgr/core/servers/deploy/install/prerequisites-for-installing-sites#primary-sites-and-the-central-administration-site):

> - If you're installing a primary site, you may also need **Administrator** permissions on additional servers. For example, where you install the initial management point and distribution point, if not on the site server.

But now I'm surprised that there's not already a link to that article in this list of prereqs. There's a link to the [Site and site system prerequisites](https://docs.microsoft.com/mem/configmgr/core/plan-design/configs/site-and-site-system-prerequisites) article, but not the installation prereqs.

There's also already a mention of what you're talking about under the final bullet:

> - Permissions for the [site system installation account](https://docs.microsoft.com/mem/configmgr/core/plan-design/hierarchy/accounts#site-system-installation-account)
>   - By default, many customers use the site server's computer account to install new site systems. The requirement is then to add the site server's computer account to the local **Administrators** group on the remote site system. If your environment uses this configuration, make sure to add the computer account of the new site server to this local group on all remote site systems. For example, all remote distribution points.

But I can see how that might get overlooked if you're not using a _site system installation account_.

So...I'm going to close this PR. I'll add some more cross-linking between these articles.

aczechowski:
@davidpaMSFT FYI I've submitted changes based on this feedback in the private repo via MicrosoftDocs/memdocs-pr#4339. I reworked the prereqs section altogether to organize them a bit better, and added better cross-linking for the other prereq articles. It will publish live Monday morning. Take a look and let me know if you have any further feedback. Thanks again!

PR #1594, opened 2021-05-07
Please modify the Docs description details

sufuruka:
Docs and actual behavior does not match.
Please update docs or modify about the conditions which user can update iOS version automatically.

Docs describes "Let" but it might be too strong words. I think it seems have something conditions other than Supervised mode to update OS automatically.
Also, Apple said Now When user need to update OS version, User need to do manually step until they set AutoUpdate settings "ON".
Apple recognized and plan to improve these behavior in the future.(*User asked to Apple Support and got this answer from Apple)

Docs notes
"Software update policies let you force supervised iOS/iPadOS devices to automatically install OS updates"
https://docs.microsoft.com/en-us/mem/intune/protect/software-updates-ios

User confirmed behavior:
Policy didn't work as they expected. Pop Up message show and need to select bottom manually to update OS.

### DIFF
--- a/memdocs/intune/protect/software-updates-ios.md
+++ b/memdocs/intune/protect/software-updates-ios.md
@@ -29,7 +29,7 @@ ms.collection: M365-identity-device-management
 
 # Add iOS/iPadOS software update policies in Intune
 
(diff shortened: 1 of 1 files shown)


PRMerger14:
@sufuruka : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Brenduns:
@sufuruka  Thanks for this edits in this PR, and the additional details surrounding that!  Its very helpful to have that larger context with a submission. 

#sign-off

PR #1609, opened 2021-05-12
Please remove the description

sufuruka:
"Although the update process doesn't typically involve any user interaction, if the device has a passcode the user must enter it to start a software update. "

The above description is not correct info.
As i told in  pull request before, Apple expressed to customer that iOS update need to user interaction. Also, this design will be change to not need to any user's interaction in the future. 
So, In current design, user need to interaction when iOS update will be done. I think that Our Docs need to modify the description.


PRMerger8:
@sufuruka : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Brenduns:
#sign-off

PR #1701, opened 2021-06-15
Note Hybrid Azure AD unsupported

CBrasga:


PRMerger16:
@CBrasga : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ErikjeMS:
Investigating with PM...

ErikjeMS:
@CBrasga This statement does not need to be a note.

CBrasga:
> @CBrasga This statement does not need to be a note.

Curious what the reasoning is? Generally exceptions are Noted. And this is an exception to "Corporate-owned Windows devices". Ego?

ErikjeMS:
We have studies that show people are ignoring Notes. Also, we try to avoid using notes except for very important things. An exception to supported platforms does not rise to this level.

CBrasga:
> We have studies that show people are ignoring Notes. Also, we try to avoid using notes except for very important things. An exception to supported platforms does not rise to this level.

Seems inconsistent. This page uses Notes to call out unsupported platforms: https://docs.microsoft.com/en-us/mem/intune/remote-actions/teamviewer-support

ErikjeMS:
Yes, it's inconsistent. Policies change over time and not all past uses of Notes (etc.) have been assessed for updating.

PR #1728, opened 2021-06-24
Update windows-10-feature-updates.md

justengraves:
I'm not certain this is absolutely accurate.  The "Feature updates" policy in Intune doesn't allow anything prior to 1909 today (unless I'm looking at something wrong).  I'm assuming that we're going to restrict "pinning" to only supported versions of the OS.  Might make sense to take the version callout out of this doc entirely and explain it as "...that you want devices to remain at.  This only includes currently supported versions of Windows 10."  Maybe even make it a purple note so that it's clearer.

### DIFF
--- a/memdocs/intune/protect/windows-10-feature-updates.md
+++ b/memdocs/intune/protect/windows-10-feature-updates.md
@@ -31,7 +31,7 @@ ms.collection: M365-identity-device-management
 
 *This feature is in public preview.*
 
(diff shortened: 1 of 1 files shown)


PRMerger17:
@justengraves : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Brenduns:
#sign-off

PR #1732, opened 2021-06-25
plural limitations

mykl-ms:
"The following are known limitation when working with shared iPads" should read "The following are known limitations when working with shared iPads"


PRMerger6:
@mykl-ms : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ktoliver:
Looks good; thanks for the fix.

PR #1733, opened 2021-06-25
RemoteSigned instead of unrestricted

discentem:
This appears to work and is more secure (I think?).


PRMerger6:
@discentem : Thanks for your contribution! The author(s) have been notified to review your proposed change.

greg-lindsay:
This is probably more secure, but we are only using a process scope anyway.  I don't see any problems with changing it.

PR #1741, opened 2021-06-30
Update vpn-settings-configure.md

roblane-msft:
When creating VPN profiles the same constraint wrt to cert based auth and using the same group to deploy the profile AND associated cert profiles applies. We describe this as a TIP in the WIFI profile but not here in VPN - customers make this mistake in the same way so would be good to have consistent message.


PRMerger15:
@roblane-msft : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger13:
@MandiOhlinger : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MandiOhlinger:
#sign-off

PR #1766, opened 2021-07-12
Update whats-new-in-version-2010.md

jasonsandys-microsoft:
Changed 2006 to 2010 per old SMSTalk reply/thread -- from March :-(


PRMerger6:
@jasonsandys-microsoft : Thanks for your contribution! The author(s) have been notified to review your proposed change.

aczechowski:
I don’t think this is correct. The first sentence is referring to functionality that we added in 2006, and then the rest of it talks about the changes in 2010.
Closing the PR. LMK if I'm missing something!

PR #1815, opened 2021-07-28
Update device-enrollment-program-enroll-ios.md

AnyaNovicheva1:
deleting the note about iPadOS 14.3.x and earlier because this was fixed with 2107, which has rolled out.


ErikjeMS:
#sign-off

PR #1834, opened 2021-08-03
Windows 11 is supported starting in 2107

PanuSaukko:
Windows 11 should be added as a supported client OS


PRMerger17:

aczechowski:
Thanks @PanuSaukko for the contribution, but that OS version isn't generally available yet, which is why we've not added it. Look for more doc updates for this in the coming months. 😉

PR #1876, opened 2021-08-11
Update windows-autopilot-hybrid.md

colealbrecht:
OU=Level 1,OU=Level2,DC=contoso,DC=com is confusing and is open to misinterpretation. When looking in AD it could be interpreted as 

```
Level 1 >
   Level2  
```

when in actuality the structure in AD for this distinguished name is:

```
Level2 >
    Level 1
```


greg-lindsay:
Looks fine

PR #1904, opened 2021-08-18
Correcting typo

danielharkerdp:
"Active AD Join" doesn't seem correct, assume it means Azure AD Join


PRMerger9:
@danielharkerdp : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #1906, opened 2021-08-18
Applies to "Windows 11" for Windows Autopilot

Yogesh-Shede:
Hi, "Windows Autopilot: What's new" link explains that applies to "Windows 11" hence below link should also get updated 
I hope my understanding is correct ?

Link > https://docs.microsoft.com/en-us/mem/autopilot/windows-autopilot


PRMerger18:
@Yogesh-Shede : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #1936, opened 2021-08-27
Update defender-advanced-threat-protection.md

pahuijbr:
- Windows Server Semi-Annual Channel (SAC), version 1803 and later

Above is the right way to refer to the version of Windows Server that gets updated twice a year like W10 (it's not 2016 rather a variant of 2019).


aczechowski (approved):
Thanks @pahuijbr for the contribution. This looks good, let's get it merged and published

PR #1991, opened 2021-09-17
The setting of region/language/keyboard should be moved

wattbt:
We had the line "Sets the region, language, and keyboard to the original values." as the first entry of the section marked by "The Windows Autopilot Reset process automatically keeps information from the existing device:" which is confusing for those reading this document (I've had some queries about it).

It makes more sense to list it in the first list of what Autopilot Reset does, leaving the second list as just those items that are kept after an Autopilot Reset.

I have also added some periods to the end of some of the list items to make them all consistent.


PRMerger12:

ktoliver:
#label:"aq-pr-triaged"

aczechowski (approved):
Thanks @wattbt for the contribution! I think this makes good sense. The author is out right now, but we'll get this merged and published shortly.

aczechowski:
#label:"ready-to-merge"

PR #2002, opened 2021-09-23
Update profiles.md

pelarsen:
Moved Language and keyboard above  **Apply device name template** as is it the order it is shown in the profile
Adding a note for Language and keyboard requiring Ethernet connectivity


PRMerger15:
@pelarsen : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #2037, opened 2021-09-29
Update resize-cloud-pc.md

abrahamp91:
Mentioned that resize is only supported for paid licenses, but support for trial licenses will come on October 11th.


PRMerger14:
@abrahamp91 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #2067, opened 2021-10-06
Updated new wording on Cloud attach wizard

wattbt:
The option "Upload to Microsoft Endpoint Manager admin center" is now called "Enable Microsoft Endpoint Manager admin center". Updated wording to reflect that.


ktoliver:
#label:"aq-pr-triaged"

PR #2131, opened 2021-10-22
Update device-rename.md

GenerAhl:
The first bullet was very confusing, like a mix between the old requirement (15 characters or less) and the new requirement (63 characters or less)


Brenduns:
#sign-off

PR #2167, opened 2021-11-01
update configuration settings for defender toggle

Priyankag16:
#sign-off

### DIFF
--- a/memdocs/intune/.vscode/settings.json
+++ b/memdocs/intune/.vscode/settings.json
@@ -1,0 +1,8 @@
+{
+    "markdownlint.config": {
+        "MD028": false,
+        "MD025": {
+            "front_matter_title": ""
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


PRMerger5:
@Priyankag16 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ktoliver:
Hi @Priyankag16 - This pull request was opened in the public repo. PMs should work in the private repo, per the Microsoft Docs contributor guide. We can keep this PR open for review and merge, but would you make future content updates in the private repo? 

The settings.json file does need to be removed from the pull request. Config files can't be merged in this public repo. Thanks.

Brenduns:
@Priyankag16   Hello Priyanka, As noted, this should be submitted via the internal repository. 
-  I will close this PR for now, with the expectation you'll submit this via the Internal repository where I can more easily work with you to make this publish-ready.    

This current PR needs a few edits to make it successful, the primary of which is to remove the  vscod/settings.json file  which is something vscode can add when it saves your local settings, and which isn't part of the file set.   Beyond that json file,  there are some case issues for references and bookmark paths, as well as a few additional minor items we should correct  before publishing these details.

PR #2241, opened 2021-11-23
Update advanced-threat-protection-configure.md

juqassis:
When clicking on the "open the Microsoft Defender Security center" link in the Endpoint Manager Portal, the URL is securitycenter.windows.com

However it looks that URL now redirects to the M365 Defender console instead. As such, the UI has changed and the path to navigate to the "Microsoft Intune connection" setting is slightly different than what is described in this article. This caused some confusion for me and my customer, so I am proposing a change with the correct path.


PRMerger18:
@juqassis : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Brenduns:
@juqassis   Thanks for this submission.  

#sign-off

PR #2255, opened 2021-12-01
Apply Operating System Image task behavior

jonasatgit:
During testing we saw the ConfigMgr Cache still has its content after the Apply Operating System Image step has run.  If that is by design, it should be part of the documentation.


jonasatgit:
I made a mistake. It is not the CCMCache folder, it is the SMSTasksequence folder which will not be deleted. Which makes sense and my pull request is obsolete.

PR #2276, opened 2021-12-08
Azure Sign-In account prerequisites.

aendrawos:
added that user needs to be an Intune admin or global admin


PRMerger-2:
@aendrawos : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Brenduns:
#sign-off

PR #2318, opened 2021-12-30
Change name of Reporting configurator

aidenmitchell:
It seems that the utility name has changed.


PRMerger18:
@aidenmitchell : Thanks for your contribution! The author(s) have been notified to review your proposed change.

aczechowski (approved):
Thanks @aidenmitchell for the edit, and _thank you_ for your first contribution to memdocs! 🎉 I think this probably changed at some point in a later version of SQL Server, so there might still be earlier versions with the "Reporting Services" name, but I think it's ok to update the string to the current version.

aczechowski:
#sign-off

PR #2332, opened 2022-01-07
Corrected statement on what happens when content not cached

wattbt:
We were stating that when a cache server hadn't yet cached content clients would download content from the internet. This isn't true, that would only happen if the cache server was unavailable, so I've corrected that sentence.

I've also made a minor amendment to step 6 under this paragraph to try and give more clarity that peer content is also retrieved in parallel and to reference clients B and C in the diagram that aren't mentioned.


PRMerger19:

ktoliver:
#label:"aq-pr-triaged"

aczechowski (approved):
Thanks @wattbt for the contribution!

aczechowski:
#sign-off

PR #2421, opened 2022-02-04
Update compliance-use-custom-settings.md

GenerAhl:
Clarified how customer compliance settings will interact with CA. The main hangup was "Custom settings can also be used" which I think should be "Custom settings will also be used". It sounded like deployed custom settings could be arbitrarily used in CA decisions when in fact they are equally considered alongside the built-in ones.


PRMerger-2:
@GenerAhl : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Brenduns:
#sign-off

PR #2470, opened 2022-02-23
Update docs to reflect public preview of iOS combined app

OluchiC:
We're in public preview of the combined client for iOS, changing the docs to reflect that so customers are aware of the preview.


PRMerger19:
@OluchiC : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Brenduns:
#sign-off

PR #2562, opened 2022-03-16
Update microsoft-tunnel-prerequisites.md

amjadaljunaidi:
since the docker CE is the same as Docker EE (the only difference is that EE supported by Docker company) then we should add both versions


PRMerger8:
@amjadaljunaidi : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Brenduns:
@amjadaljunaidi   Thanks for this PR submission.  I've checked with the product team, and they have not tested with Docker EE to ensure there are no unexpected differences from CE. Therefore, we're unable to accept this PR to add Docker EE as supported.   Until there is demonstrated customer demand for EE, they have no plans to add Docker EE to the test backlog.  

I'll now close this PR as we can't accept it at this time.

PR #2580, opened 2022-03-23
Update about-client-settings - register devices

ChadSimmons:
In Cloud services, change "Automatically register" to "Automatically join"... new Windows 10 or later domain joined devices with Azure Active Directory.  This helps avoid the confusion of AAD registered devices versus joined devices.


PRMerger15:
@ChadSimmons : Thanks for your contribution! The author(s) have been notified to review your proposed change.

aczechowski:
Thanks @ChadSimmons for making the contribution, but the text of the client setting is still "register" - so those headings should all be exactly the same as what's in the console. I don't disagree that it would be good to change it, but as long as the product has that string I'd like to keep the doc the same. Especially in case anyone searches for it. I suggest you file a frown from the console to provide feedback to the engineering team. https://docs.microsoft.com/mem/configmgr/core/understand/product-feedback

PR #2598, opened 2022-03-25
Add note to computer maintenance section.

bryandam:
This comes up from time to time.  People look at these two settings and say "oh no, my users are going to prevent me, the admin, from running deployment on their devices."  This is followed almost immediately by "how do I manage/disable these? Why isn't there a client settings for this".

Feel free to do a better job of explaining that this is both by design and will not prevent required installations from running at their deadline.  I tried finding a more pleasing way to phrase this but ran out of talent quickly as so often happens.

### DIFF
--- a/memdocs/configmgr/core/understand/software-center.md
+++ b/memdocs/configmgr/core/understand/software-center.md
@@ -197,6 +197,9 @@ Specify how Software Center applies changes to software before the deadline.
 
 - **Suspend Software Center activities when my computer is in presentation mode**: This setting is enabled by default.
 
+> [!NOTE]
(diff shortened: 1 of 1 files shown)


aczechowski (approved):
Thanks @bryandam for the contribution (trifecta for the week!) I think what you have is well worded, works for me. 😁

bryandam:
Got the hat trick!

99% of my Reddit replies have at least one link to the docs.  The other 1% tend to lead to docs PRs to remedy that unconscionable situation.  I guess this week when it rains it pours.  Remember kids: the only difference between someone raving incoherently online and someone speaking truth is a docs link.

PR #2600, opened 2022-03-25
Pavithir patch 1

Pavithir:


PRMerger9:
@Pavithir : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ErikjeMS:
I've rearranged the end user content into the parent doc set and put most of these changes into https://docs.microsoft.com/en-us/windows-365/get-users-started article.

PR #2614, opened 2022-03-31
Update boundary-groups-management-points.md

CaptMilo:
There was an extraneous 'that' included in the sentence that didn't make sense.


aczechowski (approved):
Thanks @CaptMilo for catching that, and _thank you_ for your first contribution to memdocs! 🎉

PR #2618, opened 2022-04-01
Update certificate-connector-install.md

mardahl:
added info about .NET framework version used and how to pre-install to avoid errors that don't make sense initially.
Have come across this on more than one occasion, so just adding it here for clarity.

#ATCP


Brenduns:
#sign-off  
@mardahl   Thanks for this great tip!

PR #2654, opened 2022-04-09
Consider changing the word "must" to "should"

byteben:
Setting the Feature Update deferral value to 0 in the Update ring settings is favourable and avoids unnecessary delays or complexities when targeting the same device with a Feature Update Policy. If the device is targeted with a Feature Update Policy and the Feature Update Deferral value in the Update Ring is set to a value higher than 0 then the feature update won't be offered to the device until the deferral period (in days) has been reached.

It is my understanding that the value does not "have" to be 0 but it is best practice to set this to 0. The use of the word "must" is ambiguous and suggests that the value must be zero for the Feature Update to be offered to the device, via the Feature Update policy, to be effective.

The Tip below the suggested change, in the doc, goes part way to explain the above to the reader

### DIFF
--- a/memdocs/intune/protect/windows-10-feature-updates.md
+++ b/memdocs/intune/protect/windows-10-feature-updates.md
@@ -95,11 +95,11 @@ The following are prerequisites for Intune's Feature updates for Windows 10 and 
 ## Limitations for Feature updates for Windows 10 and later policy
 
 - When you deploy a *Feature updates for Windows 10 and later* policy to a device that also receives an *Update rings for Windows 10 and later* policy, review the update ring for the following configurations:
-  - The **Feature update deferral period (days)** should be set to **0**.
+  - We recommend setting the **Feature update deferral period (days)** to **0**. This configuration ensures your feature updates are not delayed by update deferrals that might be configured in an update ring policy.
   - Feature updates for the update ring must be *running*. They must not be paused.
 
   > [!TIP]
-  > If you're using feature updates, we recommend you end use of deferrals as configured in your update rings policy. Combining update ring deferrals with feature updates policy can create complexity that might delay update installations.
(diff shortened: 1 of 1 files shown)


PRMerger18:
@byteben : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Brenduns:
@byteben  Thanks for this feedback. I suspect your suggestions is right. However, I do need to get confirmation from the  PM, the original source of that statement before I go changing it. I now suspect that rewording this to state this as a recommendation and add why (to avoid conflicts/delays due to deferrals that are set in a Ring policy) would be a further improvement.   Once I know more, I'll update you here.

PRMerger17:
@Brenduns : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Brenduns:
@byteben  After feedback from the PM, we agree this is a great clarification and we should note this as a recommendation to configure it as zero, not a requirement. I've edited it to note we do recommend the configuration, and with any recommendation, we should provide a bit of the why.  

Thanks for catching that bit and helping improve the docs!

#sign-off

byteben:
Thanks Brent. I am in exact agreement with the sentiment and change. Appreciate your help on this PR, thank you! :)

Ben Whitmore

Senior Cloud Consultant
Microsoft MVP – Enterprise Mobility
Web:https://cloudway.com<https://cloudway.com/>
***@***.******@***.***>
Mob:+447595973560<tel:+447595973560>
________________________________

PRMerger12:
@byteben: I'm sorry - only the author of this article, @brenduns, can sign off on your changes. But we do have an exception process - if you are on the Microsoft content or product team for this product area, you can ask the PR review team to review and merge it by sending mail to the techdocprs alias.

Brenduns:
@byteben   Hey Ben, I had to delete that last comment. It showed up with some of your personally identifiable information including phone number, etc.  It wasn't clear if you intended to have that added, and I thought I was only removing some specific info, but it seems I removed the entire comment. :( Sorry about that. But I appreciate your efforts to submit the PR, and the comments on the extra edit.

Issue #2613, opened 2022-03-31
Defender for Endpoint / Intune - Mismatch in docs


mikerobyn:
Hi everyone,

in the docs it says:
Select Endpoint security > Microsoft Defender for Endpoint, and then select **Open the Microsoft Defender Security Center**.
There is no such link in the portal.

Cheers,
Mike

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 24fa182f-ace6-e7e2-ff98-91951be7d10e
* Version Independent ID: f1acb398-6d69-b91c-5ee0-3da3bbf7809f
* Content: [Configure Microsoft Defender for Endpoint in Microsoft Intune](https://docs.microsoft.com/en-us/mem/intune/protect/advanced-threat-protection-configure)
* Content Source: [memdocs/intune/protect/advanced-threat-protection-configure.md](https://github.com/MicrosoftDocs/memdocs/blob/main/memdocs/intune/protect/advanced-threat-protection-configure.md)
* Service: **microsoft-intune**
* Sub-service: **protect**
* GitHub Login: @Brenduns
* Microsoft Alias: **brenduns**

msbemba:
@mikerobyn Thank you for your feedback. It is now called Microsoft defender for endpoint admin console. We have made changes to the document via PR#https://github.com/MicrosoftDocs/memdocs/pull/2688

Once the author approves, the changes will be live.

Thanks
Sri

yogkumgit:
@msbemba

msbemba:
@yogkumgit The change is accurate as far as what I see in the tenant. So I have reopened the pull request after discussing it with the author. 

Thanks
Sri

yogkumgit:
@msbemba Thank you very much for the contribution and updating the documentation with PR. @mikerobyn Hope this update is helpful for you. Thanks for taking out some time to open the issue. Appreciate and encourage you to do the same in future also.

PR #2688, opened 2022-04-13
Update advanced-threat-protection-configure.md

msbemba:
Made changes to the navigation as it is not Microsoft defender for endpoint admin console.

### DIFF
--- a/memdocs/intune/protect/advanced-threat-protection-configure.md
+++ b/memdocs/intune/protect/advanced-threat-protection-configure.md
@@ -50,7 +50,7 @@ You only need to enable Microsoft Defender for Endpoint a single time per tenant
 
 1. Sign in to the [Microsoft Endpoint Manager admin center](https://go.microsoft.com/fwlink/?linkid=2109431).
 
-2. Select **Endpoint security** > **Microsoft Defender for Endpoint**, and then select **Open the Microsoft Defender Security Center**. 
+2. Select **Endpoint security** > **Microsoft Defender for Endpoint**, and then select **Open the Microsoft Defender for Endpoint admin console**. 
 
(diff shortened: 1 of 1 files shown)


msbemba:
Issue#https://github.com/MicrosoftDocs/memdocs/issues/2613

Brenduns:
#hold-off 

@msbemba   This PR is not approved.   Here is a capture from the live Tenant this morning. The link title is "Open the Microsoft Defender Security Center"   Note this is very far down the page.    If this link doesn't appear for the customer, they should open a support incident to investigate why their tenant does not present this.   

image

msbemba:
Thank you @Brenduns .I checked in a couple of tenants and I see the below

msbemba:
I was under the impression that what I see currently is the latest as it was previously called the Microsoft Defender Security Center( as updated in the document).

Brenduns:
@msbemba  Interesting.   I'm pinging the dev/PM about this.      Stay tuned.  Meanwhile: Does your tenant already have the connection to Defender set up?  It appears to - though I don't have that set up myself  and a loaner from the Dev isn't connecting today.  I wonder if the pre/post connection to MDE affects that link text in the UI.  While  not delving into troubleshooting (which we need to avoid doing), I wonder if its a change in text label seen before/after that connection to Defender is configured. I also wonder if they just didn't see it as its ridiculously far down the page.  If they saw the link, (either form), they might have mentioned the link they do see.

msbemba:
@Brenduns That is correct and I have confirmed with a couple of other tenants and I see the same result.

JohanFreelancer9 (approved):

Brenduns:
#sign-off 

It seems MDE has some changes that haven't filtered to all content or been generally discussed in docs (that I've found).  Typically, the Security Center is has been a Discrete console used for all Security Operations, with a change being that the Defender for Endpoint admin console seems to now be co-located in that Security Center.  The standalone console for MDE was deprecated and there are now link redirections in place.

PR #2786, opened 2022-05-17
Update uninstall-applications.md

davidpaMSFT:
I don't see why we link to "App approval" section from here, which doesn't mention implicit uninstall at all. We should link either to main ("Create application groups") article, or Deploy section of that article.


aczechowski (approved):
Thanks @davidpaMSFT for catching that

PR #2937, opened 2022-07-15
Added info about Location Redirection

spatnaik:
Added info about Location Redirection and fixed some typos


PRMerger19:
@spatnaik : Thanks for your contribution! The author(s) have been notified to review your proposed change.

JohanFreelancer9 (approved):

spatnaik:
@erikjsms: Question on "Do not" to "Don't": The actual name of the Group Policy in the OS has "Do not". Do we still change the doc to say "Don't" even though it doesn't match what's in the product?

PR #2974, opened 2022-07-27
Intune Licensing requirements for AVD multi-session VMs

robgrame:
It's not clear if there any Intune licensing requirements when enrolling AVD multi-session VMs in Intune.
I'm not proposing an actual change but I believe we should provide more information in relation to licensing requirements.
We are stating that enrollment takes place using a GPO leveraging Device credentials and thus bypassing Intune license requirements for user.
While in the co-management scenario the ConfigMgr license is covering also Intune enrollment, with Hybrid Azure AD Joined-only scenario is not clear whether the user has to be assigned an Intune license.


PRMerger17:
@robgrame : Thanks for your contribution! The author(s) have been notified to review your proposed change. @smbhardwaj

PRMerger9:
@Smritib17 : Thanks for your contribution! The author(s) have been notified to review your proposed change. @smbhardwaj

Smritib17:
Based on feedback from the PM I have edited the file to include a link to the Microsoft Intune Licensing page and added a text snippet.

PR #3068, opened 2022-08-23
Document is out of date

darena-patrick:
The apple push certificates portal no longer downloads the certificate as .PEM ... it now downloads as .CER and needs to be renamed to .PEM in order to circumvent this check: 

image


lenewsad:
Hi @darena-patrick, thanks for your contribution and bringing this to our attention. We tried to reproduce what you're experiencing but the certificate file is downloading as expected in PEM format. We recommend reaching out to [Apple Support](https://support.apple.com/en-us/HT208643) to find out why the file is downloading in a different format and if that's expected behavior. I didn't add the full dedicated note, but I did add additional context (i.e. file needs to be in PEM format when uploading) to the step where the upload occurs so that people know the format requirements.

lenewsad:
#sign-off

darena-patrick:
for reference: 
image

JohanFreelancer9 (approved):

yogkumgit:
@tiburd Please merge this PR. Thanks

Issue #2969, opened 2022-07-25
Support for Windows 10 LTSC


enfo:
This page does not mention support or lack thereof for Windows 10 LTSC versions.
I know that in practice LTSC works with Autopilot and here https://docs.microsoft.com/en-us/mem/autopilot/self-deploying#requirements the documentation states that LTSC 2019 is not supported for self-deploying mode, which can be understood to mean that Auopilot in general works with LTSC 2019 but just not the self-deplying mode.

It would be very helpful to have some clear statement in regards to the LTSC versions on this requirements page.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: ca6f1187-d749-6db2-d46f-1d1d60bbc289
* Version Independent ID: f8154992-4a7a-c0ff-5cd6-c691f7fa26d3
* Content: [Windows Autopilot software requirements](https://docs.microsoft.com/en-us/mem/autopilot/software-requirements)
* Content Source: [memdocs/autopilot/software-requirements.md](https://github.com/MicrosoftDocs/memdocs/blob/main/memdocs/autopilot/software-requirements.md)
* Product: **w10**
* GitHub Login: @aczechowski
* Microsoft Alias: **aaroncz**

aczechowski:
Hi @enfo I double-checked with the PM to confirm: LTSC builds aren't currently tested with Autopilot, thus aren't officially supported, which is why they're not listed in this article. I suggest filing [product feedback](https://feedbackportal.microsoft.com/feedback/forum/ef1d6d38-fd1b-ec11-b6e7-0022481f8472).

PR #3069, opened 2022-08-23
Add  Windows 10 Enterprise LTSC to supported os

JayRHa:


aczechowski:
Thanks @JayRHa for suggesting the edit (and thank you for your first contribution to memdocs 🎉), but LTSC isn't currently supported with Autopilot, see https://github.com/MicrosoftDocs/memdocs/issues/2969

PR #3214, opened 2022-10-11
Updating Blocking personal Windows devices section

yushinok:
Updating Blocking personal Windows devices section given the investigation result of ICM (335054285).


yushinok:
I should have updated different repo.

PR #3266, opened 2022-10-17
Update set-conditional-access-policies.md

kmsft:
Searching for 'cloud' returns no results, so it should be removed.   I don't see a scenario where we would not call it Windows 365


JohanFreelancer9 (approved):

ErikjeMS:
@kmsft Please recreate this PR in the private repo. Closing.

PR #3305, opened 2022-10-25
Fix "Fast user switching" setting to the correct CSP.

reiikei:
Start/HideSwitchAccount CSP corresponds to "Switch Account" setting, while "Fast user switching" setting seems to use WindowsLogon/HideFastUserSwitching CSP.


JohanFreelancer9 (approved):

MandiOhlinger (approved):

MandiOhlinger:
@reiikei Good find. And, you are correct. Thank you for fixing this.

Issue #3326, opened 2022-11-01
Page should mention that an Intune License is required in order to set the MDM Authority.


perldj:
This page should mention that an Intune license is required by the person setting the MDM authority to be able to set the MDM Authority. I found that there was a strange error appearing when I was trying to set the MDM authority and it wasn't clear but the reason was that my Office 365 user did not have an Intune license applied, so I was unable to see the page to set the MDM Authority, even though the prompt appeared.

---
#### Document Details

⚠ *Do not edit this section. It is required for learn.microsoft.com ➟ GitHub issue linking.*

* ID: a76fc834-999b-2dea-10b7-984225b695dd
* Version Independent ID: 583b8b02-5c7b-aa08-ec28-07b89adbd994
* Content: [Set the mobile device management authority - Microsoft Intune](https://learn.microsoft.com/en-us/mem/intune/fundamentals/mdm-authority-set)
* Content Source: [memdocs/intune/fundamentals/mdm-authority-set.md](https://github.com/MicrosoftDocs/memdocs/blob/main/memdocs/intune/fundamentals/mdm-authority-set.md)
* Service: **microsoft-intune**
* Sub-service: **fundamentals**
* GitHub Login: @Smritib17
* Microsoft Alias: **smbhardwaj**

msbemba:
@perldj Thank you for your feedback. We have made changes to the document via PR#https://github.com/MicrosoftDocs/memdocs/pull/3363

Once the author approves, the changes will be live.

Thanks
Sri

msbemba:
@yogkumgit The PR has been merged and this can be closed.

yogkumgit:
@msbemba Thank you very much for the contribution and updating the documentation with PR. @perldj Hope this update is helpful for you. Thanks for taking out some time to open the issue. Appreciate and encourage you to do the same in future also.

PR #3363, opened 2022-11-15
Update mdm-authority-set.md

msbemba:
Updated the document to reflect that the admin needs an intune license to set the MDM Authority


msbemba:
Closes #3326

Smritib17 (approved):

PR #3377, opened 2022-11-21
corrected Azure subscription requirement

kmsft:
please review ASAP.  Customers are getting confused around whether there is a need for an Azure subscription.


kmsft:
edited in public instead of private

PR #3562, opened 2023-02-02
Update reusable-settings-groups.md

crosoriov:
Adding limit for number of instances per group + limit for reusable groups per Device Control profile after internal discussion with Laura Arrizza


Brenduns:
#sign-off

JohanFreelancer9 (approved):

PR #3610, opened 2023-02-23
Updated requirements

jonasatgit:
I changed the IIS requirement part to be clearer about the IIS requirements for PKI certificates and add more context even though there is a hint next to it.


jonasatgit:
I missed one sentence. Will create a new pull request.

PR #3624, opened 2023-02-28
Update resize-cloud-pc.md

ronmartmsft:
A vCPU is not a core, even though the term is misused frequently. Removed all references to core from the doc to reflect reality.


JohanFreelancer9 (approved):

ErikjeMS:
#sign-off

PR #3654, opened 2023-03-07
Update create-azure-network-connection.md

benoitms:
Owner role is required to set up the ANC in the Intune portal (as proposed with the remediate blade in Intune when the admin user does not have enough privileges to setup the ANC)


JohanFreelancer9 (approved):

ErikjeMS:
@benoitms According to Femi, we no longer require the owner role to create an ANC. Therefore, closing this PR.

Also, as a MSFT, please use the private branch for your changes in the future.

PR #3677, opened 2023-03-15
Update pre-provision.md

EleO-ms:
customer reported that they don't see the AAD sign-in prompt user flow for pre-provisioned hybrid devices which is by design (there's no need to ask for the AAD creds in PP hybrid flow and it was functional requirement from day 0 to skip that AAD sign-in during OOBE for hybrid). Adding note to that effect in the request.


JohanFreelancer9 (approved):

frankroj:
#sign-off

Issue #3689, opened 2023-03-21
Alert Type options outdated in documentation


danbu:
https://learn.microsoft.com/en-us/mem/intune/configuration/ios-device-features-settings#settings-apply-to-automated-device-enrollment-supervised)

Alert type: When devices are unlocked, choose how the notification is shown. Your options:

    None: No notification is shown.
    Banner: A banner is briefly shown with the notification.
    Modal: The notification is shown and users must manually dismiss it before continuing to use the device.

should be based on the "Configure instance" menu options

Alert type: When devices are unlocked, choose how the notification is shown. Your options:

    None: No notification is shown.
    Temporary Banner: A banner is briefly shown with the notification.
    Persistent Banner: The notification is shown and users must manually dismiss it before continuing to use the device.

Also "Add Notifications" screenshot is out of date should be "Edit instance"

---
#### Document Details

⚠ *Do not edit this section. It is required for learn.microsoft.com ➟ GitHub issue linking.*

* ID: 1a7b8e75-3872-6bbd-4d9b-b26d1eb11f81
* Version Independent ID: 74769948-8c98-19de-df74-2fcf1d399387
* Content: [iOS/iPadOS device feature settings in Microsoft Intune](https://learn.microsoft.com/en-us/mem/intune/configuration/ios-device-features-settings)
* Content Source: [memdocs/intune/configuration/ios-device-features-settings.md](https://github.com/MicrosoftDocs/memdocs/blob/main/memdocs/intune/configuration/ios-device-features-settings.md)
* Service: **microsoft-intune**
* Sub-service: **configuration**
* GitHub Login: @MandiOhlinger
* Microsoft Alias: **mandia**

msbemba:
@danbu Thank you for your feedback. We have made changes to the document via PR#https://github.com/MicrosoftDocs/memdocs/pull/3707

Once the author approves, the changes will be live.

Thanks
Sri

danbu:
Those values are different depending if you are in the "Settings catalog" or using "Templates" in configuration profiles
Setting Catalog uses "Temporary Banner" and "Persistent Banner" under "User Experience"/"Notifications"/"Alert Type" where "Templates" uses "Banner" and "Modal" under "Alert Type"

yogkumgit:
@msbemba

msbemba:
[...]

PR #3707, opened 2023-03-24
Update ios-device-features-settings.md

msbemba:
Made changes to  **Alert type**: When devices are unlocked, choose how the notification is shown.

Fixes#https://github.com/MicrosoftDocs/memdocs/issues/3689

### DIFF
--- a/memdocs/intune/configuration/ios-device-features-settings.md
+++ b/memdocs/intune/configuration/ios-device-features-settings.md
@@ -219,8 +219,8 @@ When you assign the policy to an iPhone, the page looks similar to the following
     - **Show on Lock Screen**: **Enable** shows app notifications on the device lock screen. **Disable** prevents the app from showing notifications on the lock screen.
     - **Alert type**: When devices are unlocked, choose how the notification is shown. Your options:
       - **None**: No notification is shown.
-      - **Temporary Banner**: A banner is briefly shown with the notification.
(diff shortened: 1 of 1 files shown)


msbemba:
@JohanFreelancer9 Please copy review

JohanFreelancer9 (approved):

yogkumgit:
@MandiOhlinger Please review and Signoff this PR. Thanks

MandiOhlinger:
@msbemba The original values in the article are correct and these values are shown in the Intune UI. Your changes updated the setting names, which would make this article wrong. Maybe your changes belong in the descriptions, not the actual setting names.

msbemba:
> @msbemba The original values in the article are correct and these values are shown in the Intune UI. Your changes updated the setting names, which would make this article wrong. Maybe your changes belong in the descriptions, not the actual setting names.
> 
> In its current state, this PR can't be merged.

Thank you @MandiOhlinger I was under the settings catalog and could see these options. I have added you to the issue where the user is reporting a similar instance.

MandiOhlinger:
@msbemba So, we don't document the settings in the Settings Catalog, as there are thousands of them. You can update the individual descriptions to state something like:

- This setting might also be known as **Temporary Banner**. 
- This setting might also be known as **Persistent Banner**.

MandiOhlinger (approved):

PR #3774, opened 2023-04-13
Update Autopilot "Profile status" behavior

aendrawos:
If the AP profile is assigned to a specific created group, it will change from "Unassigned" to "Assigning" and finally to "Assigned."

But if assigned to "All Devices" the "Profile status" is changing immediately to "Assigned"


frankroj:
Thank you for the contribution and feedback @aendrawos. I don't think the details of whether or not it goes into **Assigning** or not are very important, or that it is relevant. The main thing to understand is that it goes from **Unassigned** to **Assigned**. However, I do think that looking at the properties of the device and making sure that **Date assigned** is relevant. I am going to reword your contribution, dropping the information about not going into **Assigning**, but include the information about the **Date assigned** field.

frankroj:
#sign-off

PR #3786, opened 2023-04-17
Add note about ADE

markstan:
Feature requires ADE supervised enrollment


Brenduns:
Reviewed with PM/Dev and Mark (Submitter). We won't be accepting this PR as it's been determined that the source of details for the submission were no longer applicable.

Issue #3811, opened 2023-04-24
intune-portal pre-installation commands require updating with sudo/sudo -s


UsersHaveNames:
On a fresh Ubuntu 22.04 installation, the issue was in the below one liner from the documentation:

`curl https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor > microsoft.gpg`

The above one-liner has to be updated in the docs IMO, otherwise it only succeed. The user must use ` sudo -s` or prefix it with `sudo`.

NOTE: For some reason sudo is no longer needed once this has been done at least once on the OS.

---
#### Document Details

⚠ *Do not edit this section. It is required for learn.microsoft.com ➟ GitHub issue linking.*

* ID: 80924804-c247-d3bb-2245-fd3db3f3bd62
* Version Independent ID: 80924804-c247-d3bb-2245-fd3db3f3bd62
* Content: [Get the Microsoft Intune app for Linux](https://learn.microsoft.com/en-us/mem/intune/user-help/microsoft-intune-app-linux)
* Content Source: [memdocs/intune/user-help/microsoft-intune-app-linux.md](https://github.com/MicrosoftDocs/memdocs/blob/main/memdocs/intune/user-help/microsoft-intune-app-linux.md)
* Service: **microsoft-intune**
* Sub-service: **end-user**
* Product: ****
* Technology: ****
* GitHub Login: @lenewsad
* Microsoft Alias: **lanewsad**

UsersHaveNames:
Apologies, things started working. The issue was in the below one liner:

`curl https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor > microsoft.gpg`

The above oneliner has to be updated in the docs IMO, otherwise only succeed if the used first enters `sudo -s`

lenewsad:
Hi @UsersHaveNames @yogkumgit, thanks for reaching out to the Microsoft Intune docs team and providing this info. I'm checking with the product team about PR #3816, and shared the info from this issue with them for context. I'll follow up with more info when I hear back.

lenewsad:
Closing per [PR 3816(https://github.com/MicrosoftDocs/memdocs/pull/3816).

PR #3816, opened 2023-04-25
Update microsoft-intune-app-linux.md

UsersHaveNames:
adds sudo to chained curl commands since otherwise they fail on a fresh Ubuntu installation

related to https://github.com/MicrosoftDocs/memdocs/issues/3811

### DIFF
--- a/memdocs/intune/user-help/microsoft-intune-app-linux.md
+++ b/memdocs/intune/user-help/microsoft-intune-app-linux.md
@@ -45,7 +45,7 @@ Run the following commands in a command line to manually install the Intune app 
 2. Install the Microsoft package signing key.  
 
    For Ubuntu 20.04:  
-    `$ curl https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor > microsoft.gpg`  
+    `$ sudo curl https://packages.microsoft.com/keys/microsoft.asc | sudo gpg --dearmor > microsoft.gpg`  
  
     `$ sudo install -o root -g root -m 644 microsoft.gpg /usr/share/keyrings/` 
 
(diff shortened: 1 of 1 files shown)


UsersHaveNames:
NOTE: For some reason sudo is no longer needed once this has been done at least once on the OS.

JohanFreelancer9 (approved):

lenewsad:
Hi @UsersHaveNames, thanks for your contribution to the Microsoft Intune docs! Thanks for your patience while I verify that this change is needed with the product team.

memdocs/intune/user-help/microsoft-intune-app-linux.md
```diff
@@ -45,7 +45,7 @@ Run the following commands in a command line to manually install the Intune app
 2. Install the Microsoft package signing key.  
 
    For Ubuntu 20.04:  
-    `$ curl https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor > microsoft.gpg`  
+    `$ sudo curl https://packages.microsoft.com/keys/microsoft.asc | sudo gpg --dearmor > microsoft.gpg`
```
benferse, line 48: The only reason I can see that you'd need elevated permissions in here is if you were trying to redirect the output of `gpg` to somewhere you didn't have permissions to write. e.g. you're trying to redirect the output of `gpg` to a file in a directory where you don't have permissions to create a file.

As a general rule I'd say that's implied or out of scope for the elevation of these docs, but if you want to suggest a change that uses an absolute path for `microsoft.gpg` that is more likely to be world writable, perhaps `$HOME/microsoft.gpg`?

benferse (changes requested):

memdocs/intune/user-help/microsoft-intune-app-linux.md
UsersHaveNames: You would know the best approach for sure. I can only speak as an engineer that enrolled Linux devices using Intune for the first time.

As funny as it may sound, having done this the first time, on a fresh ubuntu installation the described exception occurs. It took me a few hours before I decided to stop investigating if the issue is in dependencies indeed (which it was not) and check two commands above before I realized that intune dependencies could not be installed specifically because of this.

My advice is only to improve user experience by ensuring that whatever commands are presented, work with clean default installations of ubuntu 22.04.

lenewsad:
Hi @UsersHaveNames, thanks again for opening this PR and being willing to chat with us about your experience. At this time, we're not going to make any doc changes, but I am seeing the benefit in having an FAQ or troubleshooting guide that addresses topics like yours. I'll work with the product to create a new support topic by summer. 

Please don't hesitate to open more GitHub issues or PRs for us in the future.

PR #3846, opened 2023-05-03
Update sql-performance.md

FranciscoNabas:
The default retention period for change tracking on MS SQL Server is 2 days, but for the configuration manager, it should always be 5 days. The places where I got this info from:
 - When installing a site server, the database auto cleanup retention period is 5 days.
 - When looking the code for the stored procedure 'spDiagChangeTracking', while gathering change tracking information we can see a comment saying that for SCCM it should always be 5 days.
 
image

 - While troubleshooting a change tracking issue, I've set the retention for two days, and the auto cleanup couldn't keep up.

I understand that while troubleshooting performance issues, cleaning the change tracking information up to 2 days can improve performance, however this can be misleading, since this query will always return 'there is change tracking backlog, please contact Microsoft support'. Cleaning more change tracking than default can also mask the real cause for performance issues.

Thank you!


Banreet (approved):

Banreet:
#sign-off

JohanFreelancer9 (approved):

PR #3999, opened 2023-06-30
Clarifying LTSC in Update windows-10-update-rings.md

DavGuy-MSFT:
Making it clearer that QU is supported but Feature Updates are not


Brenduns:
#sign-off

DavGuy-MSFT:
Thank you.

PR #4017, opened 2023-07-10
Update azure-firewall-windows-365.md

jurasmus:
Same rulename is not allowed in Azure Firewall.
Giving it numbers fixes this


JohanFreelancer9 (approved):
Copy review.

ErikjeMS:
checking with pm femi

ErikjeMS:
The PM responded thus: Rule Description is purely to outline what it's for, not for use as a label. The customer can create whatever rule name they want when creating it. We don't specify in the table what the rule name should be and leave that to the customer. 

Therefore, we won't make this change.

PR #4082, opened 2023-08-01
Update device-passcode-reset.md

chriti:
Added note to document the requirement for setting a passcode configuration within device restrictions.


chriti:
Added note that the Device Passcode must be configured in the Device Restriction policy or the passcode reset will fail.

Smritib17:
#sign-off

PR #4102, opened 2023-08-10
Update tenant-status.md

schenardie:
We had some discrepancies on some tenants on number of user with the Intune SKU enabled vs Total count of licensed users. Upon opening a ticket with Microsoft the reply was that independent of the Intune SKU being enabled or disabled this count still include as one license on the "Total licensed users" under Tenant Status.


Brenduns:
@schenardie   Thanks for this PR. Clear, and direct. Much appreicated.

Brenduns:
#sign-off

PR #4121, opened 2023-08-21
Update apps-win32-add.md

aendrawos:
I had to add the word "Only", as some clients and CSS were confused and thought that the device will eventually reboot, but not force the user to do it now.

By adding only, we clearly say that no restart is scheduled.


Erikre (approved):

Erikre:
Thanks for the contribution! #sign-off

yogkumgit:
@tiburd Please merge this PR. Thanks

PR #4122, opened 2023-08-21
Update apps-win32-app-management.md

aendrawos:
Explaining that Restart Grace period settings is ONLY honored when the device returns a code mapping to hard reboot (In case of device install behavior is set to "Determine behavior based on return codes")

OR

When the device install behavior is set to "Intune will force a mandatory device restart"


Erikre (approved):

Erikre:
#sign-off

PR #4159, opened 2023-09-06
Update certificates-pfx-configure.md

Arkert:
Fixed minimum key size typo


Brenduns:
@Arkert   Thank you for this edit/fix.  That typo suggests a ridiculously large size. :)     I’ve checked the other related entries (submitted at the same time), and this was the only edit that jumped the rails unnoticed.   


#sign-off

PR #4165, opened 2023-09-06

JeffMichelmore:
During a MSFT internal call, I learned that tamper protection is not currently supported via MDE security configuration management. I want to update the public documentation to reflect this.


Brenduns:
Validated this update via Intune PM.  

#sign-off

PR #4215, opened 2023-09-19
Update task-sequence-steps.md

treytMSFT:
Clarifying a specific situation where group policy will apply during a task sequence.


frankroj:
#sign-off

PR #4242, opened 2023-10-02
Packages and Programs Maintenance Window note

ChadSimmons:
The note says

> "If you use maintenance windows on the same collection to which you deploy this program.."

This is misleading since maintenance windows apply to the client computer, not a specific collection or deployment.

The updated language clarifies this.


JohanFreelancer9 (approved):

BalaDelli (approved):

BalaDelli:
#sign-off

dereklangMSFT:
Maintenance windows are targeted to collections. Devices inherit the maintenance window policy via the collections they are members of, but maintenance windows are not targeted directly to devices.

ChadSimmons:
> Maintenance windows are targeted to collections. Devices inherit the maintenance window policy via the collections they are members of, but maintenance windows are not targeted directly to devices.

Yes, that is much more clear than my original comment.  Still, I think the PR language clarifies the original text.  Agree?

BalaDelli:
#sign-off

dougeby:
This was signed-off on but not merged. Merging.

PR #4273, opened 2023-10-09
Update app-wrapper-prepare-ios.md with new -dl command line

tranamst:
Problem: If a wrapped app implements ADAL or MSAL, they need to add the "ADALLogOverrideDisabled"=True to their IntuneMAMSettings dictionary or else we will crash setting the log callback. MSAL logging is allowed by providing an API to log info to MSAL. This can only be done once. If app is doing it and we're doing it, then app will crash.

Solution: We can provide customers of AppWrapper a command line parameter that set the ADALLogOverrideDisabled = true in the IntuneMAMSettings dictionary so that it disable logging and avoid crashing. This PR is to add public documentation for customers about this new command line.


Erikre (approved):

Erikre:
#sign-off

PR #4290, opened 2023-10-16
Update autopilot-faq.yml

schenardie:
Add information regarding self-deploying mode coupled with local accounts where the device doesn't process apps installs until the User ESP is finished or skipped.


frankroj:
#sign-off

frankroj:
Thanks again @schenardie for the contribution. I actually accidentally signed off on the other day before reviewing, but it never published. I reviewed today and reworded the question and answer just to make the issue and resolution a bit clearer. This should go live later today.

frankroj:
#sign-off

PR #4320, opened 2023-10-23
Update automatic-intune-enrollment.md

ReuelCh:
reflect AD -> Entra ID rename


frankroj:
Thank you for the contribution @ReuelCh. However we are currently automating the change from Azure AD to Microsoft Entra throughout the docset. This should be done by the end of the month. There is no need to manually edit.

PR #4331, opened 2023-10-26
Remove editor note

nathanmcnulty:
Removed question about support for user groups left in the doc. Still would be nice to know, but likely not intended to be left in here.


lenewsad:
#sign-off

lenewsad:
Whoops, thank you @nathanmcnulty for this speedy edit. :) I will follow up with the product team to find out the answer to that question and add that info in the doc.

PR #4332, opened 2023-10-26
Update whats-new.md

SubhamV:
As per the Readiness content presented on 25 Oct, it was clarified by Mike Danoski and Kevin Sheehan that the feature is available only for Windows 11 and not for 10 as mentioned in the document. I have rectified the same.


Brenduns:
#sign-off

PR #4343, opened 2023-10-31
Update autopilot-faq.yml

schenardie:
Make more the scenario more clear. The issue is that applications will not install after ESP finished because user ESP is not processed and a device token is not generated until it is. So either going through the user esp or skipping it is required for the device token to be issued.


JohanFreelancer9 (approved):

frankroj:
Thanks for the contribution @schenardie. Can you clarify what "it" is in the statement "and a device token will not be issued until **it** is processed"? Is it the ESP?

schenardie:
Hi Frank. Correct. "it" relates to the  user portion of the esp

frankroj:
Thanks @schenardie. I have the clarification in the doc and will go ahead and approve. This should go live either later today or tomorrow.

frankroj:
#sign-off

PR #4352, opened 2023-11-02
Update filters-supported-workloads.md

minakash:
Confirmed that App and browser isolation, and Device control support Filter feature from Intune admin center.


matt-call (approved):
Looks good. Validated in Canary UX that Device Control and App/Browser isolation had filters.

PR #4369, opened 2023-11-06
Update filters-device-properties.md

RicardoGFGomes:
The statement "Devices must be Intune enrolled to use this app property" don't make sense and it's adding confusion to customer's reading this article. This filter property can indeed be used to filter a policy for a MAM scenario. Even the example given on how to use the property reflects this:  '(app.deviceManagementType -eq "Unmanaged")'.

### DIFF
--- a/memdocs/intune/fundamentals/filters-device-properties.md
+++ b/memdocs/intune/fundamentals/filters-device-properties.md
@@ -1,4 +1,4 @@
----
+Devices must be Intune enrolled to use this app property---
 # required metadata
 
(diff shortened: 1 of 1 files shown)


MandiOhlinger:
@RicardoGFGomes Hey - The article is correct. The `deviceManagementType` filter property is specific to apps, not devices. So, to use the `deviceManagementType` app filter property, the device must be managed by Intune. The `deviceManagementType` filter property is also listed under the **Managed app properties** section.

If there's something we can do to make that clearer, please create a pull request with your recommended changes. Thank you!

PR #4375, opened 2023-11-06
Update app-wrapper-prepare-ios.md with prerequisites

tranamst:
From our ICM (Incident-406091722 Details - IcM (microsofticm.com)) where customers ran into code signing issue based on not having valid certs locally, we want to update our prerequisite steps for customers to self-diagnosing and self-resolve issues before reaching out to Intune oncall engineers.


Erikre:
#sign-off

PR #4377, opened 2023-11-07
Update app-reliability.md

SKjerland:
It looks like this is consistently capitalized as "Endpoint analytics" in other articles, but there were numerous instances of "endpoint analytics" in this article that needed to be fixed, so I fixed them.


Smritib17:
#sign-off

Smritib17 (approved):

PR #4399, opened 2023-11-16
Update vpn-settings-ios.md

Kirtchayan:
Automatic VPN: iOS supports either per-app or on-demand VPN. The Intune portal allows configuring both simultaneously, leading to VPN connection issues. Adding clarification about this in the article will help IT professionals avoid encountering this problem.


MandiOhlinger (approved):

MandiOhlinger:
@Kirtchayan Thank you!

PR #4423, opened 2023-12-01
Better info about which discoveries are required and two times Entra User discovery.

PanuSaukko:
Just tested yesterday: the user must have been discovered by both Entra ID and Active Directory user discovery. Also, there was incorrectly mentioned two times: Entra User discovery when the latter one was referring to AD user discovery.


JohanFreelancer9 (approved):

BalaDelli:
##sign-off

Banreet (approved):

Banreet:
#sign-off

PR #4446, opened 2023-12-11
Update enrollment-autopilot.md

andrew-s-taylor:
Changed contains to startsWith

PR #4448, opened 2023-12-12
Update enrollment-autopilot.md

schenardie:
Replace the operand 'contains' with 'startsWith', it achieves the same purpose and its not that costly to process.


frankroj:
Thank you for the contribution @schenardie. However, we're already working through this change via the following PR:

https://github.com/MicrosoftDocs/memdocs/pull/4446

PR #4469, opened 2024-02-01
Update app-discovered-apps.md

byteben:
There may be a better way to word this but the IME uses a specific WMI class to scan for "Detected Apps" data. This is NOT limited to just MSI based apps. EXE based apps are also inventoried if they have a DIsplayName and DisplayVersion entry in the ARP registry keys


Erikre (approved):

Erikre:
Thanks! #sign-off

byteben:
This is incorrect and misses the spirit of my request :( 
It's not just MSI data that is inventoried. Anything containing a DisplayName and DisplayVersion in the ARP Unjnstall key gets inventoried. EXE apps are also inventoried with the WMI class that the IME uses to build the inventory data

Erikre (approved):

Erikre:
Thanks for the contribution! #sign-off

Erikre:
#sign-off

PR #4473, opened 2024-02-07
Update whats-new.md

Jcodedk:
Intune is not spelled InTune. Best regards MS EMS community


Brenduns:
[Jcodedk](https://github.com/Jcodedk)  Thank you Jonas.    Our team has now discussed this to ensure we have a clear and singular intent, which is to ensure that in our docs the product branding is accurate, even if it’s presented incorrectly as part of a third-party’s presentation (as in this case of this apps name).

There happens to be one other instance of *InTune*  that I found hiding in the docs set, which I’ll fix in a separate PR.  

I appreciate your bringing this capitalization to my awareness.   

#sign-off

PR #4484, opened 2024-02-21
Update registration-auth.md

brianbaldock:
Include note on new requirements for device management for partners through GDAP with link to blog article.


frankroj:
This change is being reverted back for now. I will update the docs as necessary once the change is implemented again.

PR #4488, opened 2024-03-02
Additional information for hybrid-joined devices and Intune device encryption

felpsey:
Entra ID and Active Directory hybrid joined devices must have some Group Policy configuration for BitLocker Drive Encryption to provision Silent Encryption.

By default, Active Directory joined devices will attempt to write the key to Active Directory Domain Services first when Intune settings instruct the device to backup the recovery key to Entra. However, if Group Policy does not explicitly permit this action, Silent Encryption will fail.

Example error in Event Viewer when no Group Policies have been applied:

Screenshot 2024-03-02 at 21 42 50

Additionally for standard users, the ability to generate recovery keys must also be permitted.

### DIFF
--- a/memdocs/intune/protect/encrypt-devices.md
+++ b/memdocs/intune/protect/encrypt-devices.md
@@ -170,7 +170,7 @@ Depending on the type of policy that you use to silently enable BitLocker, confi
 You must mirror settings applied in Intune and Group Policy for hybrid joined devices.
 
 1. Start Group Policy Management Console (gpmc.msc).
-2. Expand Forest > Domains > _your domain>.
+2. Expand Forest > Domains > _your domain_.
 3. Right-click _your domain_ and select **Create a GPO in this domain and link it here**.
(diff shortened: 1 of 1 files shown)


felpsey:

Brenduns:
Passed to PM/Dev for feedback on 3/4 
Bumped request, 3/12.

Brenduns:
@felpsey  Thank you for taking the time to create and provide background for this content submission.  It’s taken some time to complete, but after a review by our product team, we won’t be accepting this PR to update this content.  

After investigation, the product team has confirmed that configuration of group policies for the hybrid joined devices isn’t expected nor necessary to silently enable BitLocker encryption on those devices. The source of why this might be the case for an environment isn’t something they can identify form the details they have available.

If you find these configurations are necessary for your environment, consider reaching out to support to have the source of that in your environment investigated, which is something we can’t manage through this channel.

PR #4492, opened 2024-03-08
Update device-restrictions-android-enterprise-personal.md

simonpucher:
Zero is not possible on this setting. If you enter blank then Intune uses the default setting.


MandiOhlinger:
#sign-off

PR #4498, opened 2024-03-18
Update add-devices.md

thikei:
The paragraphs are wrongly divided.


aczechowski:
Thanks @thikei for the contribution, but I think that indentation is planned/intentional, so that it nests under step 1. If we remove the whitespace from the start of line 102, it actually breaks the ordered list - the next step ("While OOBE is running...") would become another `1` item. Here's a preview of what it would look like if this edit is accepted: 
image

So I'm going to close this PR. @frankroj can review further if he wants.

frankroj:
Agree with @aczechowski. The article and current indentation is correct.

PR #4502, opened 2024-03-20
Update device-profile-assign.md

NEllis280:
Adding a note for AVD multi session scenario.


NEllis280:
@MandiOhlinger looks great! Thank you for the assist!

PR #4507, opened 2024-03-26
Update intune-endpoints.md

mimiyano:
There is both part "Microsoft Azure Attestation part" and "Migrating device health attestation compliance policies to Microsoft Azure attestation".

It seemed Dupulicated and the each info is not same,so it makes customers comfuse.So if one part  is enough ,other part will be deleted.


Smritib17 (approved):

Smritib17:
@mimiyano thank you for the inputs. The change has been made here: https://github.com/MicrosoftDocs/memdocs-pr/pull/14680

PR #4508, opened 2024-03-27
Corrected spelling/typo

SMG0927:


Brenduns:
@SMG0927 Thanks for your sharp eyes, and for this content PR with the fix. The article should appear as fixed later today.

While _Synch_ is often seen as an acceptable but uncommon short-form of synchronization, it stands out as uncommon enough in our content to support banishing it.  I did some investigation when this landed a short while ago to confirm that this is the only instance of _synch_ in our Intune docs repository (where sync appears over 500 times).  This being the only instance lends weight to support eradicating it.

#sign-off

PR #4522, opened 2024-04-03
Logical correction to technical content

SMG0927:
'Send Push Notification to end user'  was repeated as part of the list
Instead, one of the two should be 'Send email to end user'
This list would then make logical sense for actions to take for non-compliance


lenewsad (approved):
Thank you for your contribution to the Microsoft Intune docs @SMG0927!

lenewsad:
#sign-off

PR #4525, opened 2024-04-05
Updated dead link. Updated Single App Mode to App Lock per Apple's we…

StarScreamV2x:
Updated dead link. Updated Single App Mode to App Lock per Apple's website.


MandiOhlinger:
@StarScreamV2x Thank you!

PR #4527, opened 2024-04-08
Update intune-endpoints.md

MasakiMori41:
When deploying Win32 apps, use ports 80 and 443, as this is done from the browser.


Smritib17 (approved):

Smritib17:
@MasakiMori41 Thank you for the feedback. I've verified with the engineering team and updated to include TCP 443. As we use HTTPS, the default is 443.

PR #4530, opened 2024-04-10
Update policy-sets.md

Benjamin-Testart:
Adding supported apps


Erikre:
@Benjamin-Testart - Thanks for the contribution. #sign-off

PR #4531, opened 2024-04-12
Update apps-inc-exl-assignments.md

Benjamin-Testart:
My change gives a glimpse of the new Microsoft Intune console and the new group assignment experience.


Benjamin-Testart:

Erikre:
@Benjamin-Testart - We can't include a PR that includes external resources, such as images. However, I am incorporating your suggestions into this topic. Appreciate the feedback. Closing this PR now. Thanks

PR #4533, opened 2024-04-16
fix suggested for pre-requisites and App-Locker Policy activation

Mohamed-Elwan:
We've had this requested suggested from support case with customer strongly recommending to apply these changes. So, Docs can be more specific and very clear.


Brenduns:
Review/inestigation in progress.

PR #4535, opened 2024-04-25
WinRM service and client configuration on Entra ID joined devices

niklasrast:
I want to add this to memdocs as it is missing currently. After I figured out how to configure WinRM to work on devices that are not domain joined and instead are managed by Intune and joined to Entra ID I created a post on my personal blog but thought it would be nice if the knowledge could be found right in ms learn. The described configuration can be adopted easily and worked in my test and prod environment.

As it is my first time contributing to ms learn please tell me if something has to be changed or done in another way.


niklasrast:

PR #4542, opened 2024-05-14
Update support-for-windows-adk.md

gwblok:
Better explanation as to why they can't use 22H2 (22621) ADK for ARM64 Deployments.  I was testing with the older version because it does have ARM64 WinPE and I didn't want to upgrade my CM to use the 25398 ADK with brings along a pile of other issues.

PLEASE RELEASE THE 24H2 ADK ASAP, and have CM Support it!


BalaDelli:
#sign-off

PR #4550, opened 2024-06-03
Update assign-clients-to-a-site.md about Sites - Minor change

IvyMS1990:
Button 'Find Site' should be in tab Site, not Advanced


PalikaSingh (approved):
The changes seem to be correct after checking the app from control panel hence approved.

PR #4553, opened 2024-06-05

DarCampbell:
Duplicate sentence with less description that the second sentence


Brenduns:
@DarCampbell   Thank you for this PR!  I agree with your comment. That first sentence really doesn't offer much at all. 

#sign-off

PR #4555, opened 2024-06-06
Update device-images.md

MayDay47:
Added information that an image should have never been Intune enrolled to work with W365


ErikjeMS:
checking with pm eva

PR #4560, opened 2024-06-13
Update scope-tags.md

AdrienGODET:
Scope Tag cannot by apply on devices synchronized with mecm with Cloud attach


Smritib17:
@AdrienGODET , thank you for your input. Based on feedback that I got from the engineering team, If these are devices in Intune, and they are enrolled in intune with AAD Security groups, then they can have scope tags applied.

PR #4563, opened 2024-06-18
Bdam extension docs https update

bryandam:
The ConfigMgr 2403 release fully deprecated HTTP. Now, one would have thought that only applied to the infrastructure itself but multiple console extension developers found out the hard way that it was deeper.

There _may_ be other URIs similarly affected but the http://schemas.microsoft.com/SystemsManagement is the only one I've changed here as I have proof that it needs updated.


Banreet (approved):

Banreet:
#sign-off

PR #4566, opened 2024-06-21
Update endpoint-security-account-protection-policy.md

mimiyano:
Hi.

I'd like to confirm any function is preview or GA. About this doc,under "Prerequisites for Account protection profiles",it shows   "Local user group membership"  is as preview function. But there is no word"preview" about  "Local user group membership"  on both ①"Account protection profiles" part and ②Microsoft Intune Admin center.

So could you please delete the word "preview" for Local user group membership"  under "Prerequisites for Account protection profiles" of the docs？


Brenduns:
Requested confirmation from the product team.  There hasn't been a comms cycle initiated to announce the profile has moved out of preview to generally avaiable, but its possible.

Brenduns:
@mimiyano  Thanks for this submission. I checked with the product group, and they confirmed that while this wasn't an announced change, the _Local user group membership_ profile is no longer in preview.    

#sign-off

PR #4567, opened 2024-06-21
Update configure-authentication.md

dereklangMSFT:
Removed confusing language regarding EHTTP vs. HTTPS management points and how client authentication is handled.

EHTTP does not use client auth certs from internal CA's


BalaDelli (approved):

BalaDelli (approved):

BalaDelli:
#sign-off

PR #4568, opened 2024-06-24
Update intune-endpoints.md

Takashi-kg:
The customer pointed out that adl.windows.com is missing from this Docs.  adl.windows.com is included in Following Windows update endpoint requirement as HTTPS access.  https://learn.microsoft.com/ja-jp/windows/privacy/manage-windows-11-endpoints

Also, this "adl.windows.com" is the URL used by the Windows Update Safeguards function according to the following wiki. https://www.osgwiki.com/wiki/Feature_Safeguards

So, this endpoint should also be included in Autopilot Windows Update section of this Intune endpoint requirement.


triplanedave (approved):
Looks good.

PR #4571, opened 2024-06-26
Added BitLocker encryption race condition issue

rogersoMS:


frankroj:
Thanks for the contribution @rogersoMS. However, the Known Issues article is reserved for specific known issues with Windows Autopilot device preparation itself as documented by the Windows Autopilot product group. In addition, this specific issue is more of a BitLocker issue that just happens to affect Windows Autopilot device preparation - it's not an actual issue with Windows Autopilot device preparation itself.

As a Microsoft employee, I would also encourage you to use our internal repos and tools. I will send you more info offline.

PR #4574, opened 2024-06-30
Fixed table header in step 3-2

mattisafur:
Header text was showing up twice, as header and on the first row after. Removed header from second row.


mattisafur:
@BalaDelli can this be merged?

PR #4577, opened 2024-07-09
Update apps-add-android-for-work.md

alesecurity:
Changed documentation to correctly reflect the fact that Wi-Fi requirements configuration option is not supported by Intune on personally-owned work profile devices:
There is no configuration menu in Intune to do this as the policy about changing this setting is only available in profiles that target fully managed, dedicated or corporate-owned work profile devices.


Erikre (approved):

Erikre:
#sign-off

PR #4579, opened 2024-07-11
Update company-portal-app.md

YeehomZhu:
Update website name column in table according to https://portal.microsofticm.com/imp/v3/incidents/details/490521783/home


Erikre:
#sign-off

PR #4584, opened 2024-07-15
Update remote-help-windows.md

FreeFree13:
Adding the noncompliant device conditional access steps


kara-wang:
@FreeFree13 , Hi. We have to reject this PR. We are aware of this limitation currently and working on a more sustainable fix. We cannot recommend customers to exclude Microsoft Intune and Entra from their CA policy.

PR #4595, opened 2024-07-26
Update app-protection-policies.md

Alexufo27:
Custom apps aren't able to mix with All Apps, Microsoft Apps, or Core Microsoft Apps


Erikre:
#sign-off

PR #4596, opened 2024-07-28
Update troubleshoot-windows-365-boot.md

kev-out:
added an additional troubleshooting step :
when user has multiple cloud pcs and has no default cloud pc set
 "something went wrong" will be also displayed on login before user has selected a default pc on windows365.microsoft.com 

validated these steps on Windows 11 Enterprise 23H2 - OS Build 22631.3880


ErikjeMS:
#sign-off

PR #4597, opened 2024-07-29

JeffMichelmore:
DCs of newer OS can unintentionally be managed, not just downlevel. I have internal repro of a WinServer 2022 Domain Controller that is successfully managed by MDE.

### DIFF
--- a/memdocs/intune/protect/mde-security-integration.md
+++ b/memdocs/intune/protect/mde-security-integration.md
@@ -127,7 +127,7 @@ Security settings management doesn't work on and isn't supported with the follow
 
 > [!IMPORTANT]
 >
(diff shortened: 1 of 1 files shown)


Brenduns:
Pending review with Dev/PM resources.

Brenduns:
@JeffMichelmore  After consultation with the Defender product team, we won't be accepting this PR to edit the docs.  
 
The feedback and reason I've been provided follows: 

- Thanks for verifying with us.
  This change was not expected and shouldn’t be applied to the docs.  There was an incident involving unintentionally managed DCs in general, which has been addressed and communicated to the relevant customers. Therefore, the documentation can remain as it is, and no changes are needed.

PR #4601, opened 2024-08-02
Update intune-endpoints.md

ayumishimada:
By capturing packets, we found that the endpoints I've added are needed when you install a Microsoft store app.


Smritib17:
closing this one as it is a duplicate.

PR #4603, opened 2024-08-04
Update devices-wipe.md

v-2maban:
"Retain enrollment state and user account" has been replaced with "Wipe device, but keep enrollment state and associated user account"


v-2maban:
Document has been verified

Smritib17 (approved):

PR #4604, opened 2024-08-07
Updated certificate authentication note

DickTracyII:
Updated the note on this page for the OOBE and certificate authentication. I have supported several customers with this capability. There have been inquiries about this from both customer and colleagues and I thought I could help and update a small part. 

My understanding is that this is now supported when using CBA, correct? my concern is it directly conflicts with [this doc](https://learn.microsoft.com/en-us/entra/identity/authentication/concept-certificate-based-authentication-smartcard#windows-out-of-the-box-experience-oobe. ); which I also linked it to. 

Thanks!

### DIFF
--- a/autopilot/requirements.md
+++ b/autopilot/requirements.md
@@ -105,7 +105,7 @@ Additional configuration might be required to grant access to required services 
 
 > [!NOTE]
 >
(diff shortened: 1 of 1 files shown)


frankroj:
Thanks for the contribution @DickTracyII. However, as discussed, since the PG doesn't officially support this, we won't be making this change to docs. Also remember as a Microsoft employee, you have access to the internal repo so please use that in the future.

PR #4608, opened 2024-08-10
Update resilience.md

TomHickling:
Just moving the picture and two typos


TomHickling:
Moving this to the private repo

PR #4617, opened 2024-08-22
Update create-dynamic-device-group-all-cloudpcs.md

CKunze-MSFT:
Contains is a VERY costly operator for dynamic groups. It is recommended that either eq or startsWith is used instead.


CKunze-MSFT:
changed the contains operator in the dynamic group

PR #4619, opened 2024-08-23
Update azure-ad-join-technician-flow.md

matthewjlevy:
Corrected number of next step (this is step 8 and so next step is step 9: User flow)


frankroj:
Thanks for catching this typo @matthewjlevy. It should go live later today.

frankroj:
#sign-off

PR #4622, opened 2024-09-03
Update custom-settings-linux.md

aendrawos:
The current statement is wrong

The script will only start to run after that the user gives consent.

After that consent, the script can keep executing normally.


MandiOhlinger:
@aendrawos Thank you!

PR #4626, opened 2024-09-06
Update create-dynamic-device-group-all-cloudpcs.md

JrndD:
To my knowledge we can use "Starts with" and not "Contains" which is a more performant dynamic query.


ChristianMontoya (approved):
Approved. After looking at other Entra ID docs and validating myself, 'Starts With' is more efficient and recommended.

PR #4630, opened 2024-09-10
Update multi-factor-authentication.md

JrndD:
Should mention in the article that Temporary Access Pass is a solution where a second device is not available.


lenewsad (approved):
Thank you for your contribution to the Microsoft Intune docs, @JrndD!

PR #4633, opened 2024-09-10
Update requirements.md

RicardoGFGomes:
Updating the document to make it more specific on what Device Enrollment Restriction needs to have the Allow MDM enrollment set. How it's currently written, it seems we can just enable it on any restriction policy, but if we do this the Windows 365 provisioning will fail. Instead, we need to make sure that Allow MDM is set to allow in the default policy. I also updated the Restriction Policy docs, to include the Windows 365 under the limitations (Pull request #16066, approved and published on 10/09/2024)


ErikjeMS:
#sign-off

PR #4635, opened 2024-09-12
Update intune-govt-service-description.md

mselcox:
Removed note about device query not being supported in GCC-H because it is now supported.


MandiOhlinger:
@Smritib17 Can you confirm that Device Query supports GCC High?

MandiOhlinger:
#sign-off

PR #4640, opened 2024-09-26
Fixed prereq permissions needed

RyFellows:
Removed the Intune Administrator role since it does not have permissions to change Automatic Enrollment settings.


lenewsad (approved):
Thanks for your contribution @RyFellows!

PR #4643, opened 2024-09-30

andrewjohnporter:
Changed typo on OS from 2109 to 2019


Brenduns:
@andrewjohnporter   Thank you for this PR. We had some quickly turned around content to update the Server Core details in this article, that clearly went wrong. That fail is on me, so again, my thanks for submitting that fix.

#sign-off

PR #4645, opened 2024-10-05
Update reports-export-graph-available-reports.md

byteben:
The AppInvRawData endpoint does accept a filter using the eq operator.

AppInvAggregate report also returns "Platform" which is omitted from this doc


Erikre:
#sign-off

byteben:
Thanks!

PR #4646, opened 2024-10-07
Remove note about CM 2006

lamosley:
CM 2006 and earlier has been out of support for more than 3 years; TP updates are enabled by default now, so we should remove this note.


BalaDelli (approved):
Approved

BalaDelli:
#sign-off

PR #4650, opened 2024-10-14
Fixing formatting issue on line 324

brookd2404:
Line 324 was missing ** before Derived Credentials leaving the web page showing the last two starts and not marking it as bold


Brenduns:
@brookd2404  Thanks for submitting this fix! 

#sign-off

PR #4654, opened 2024-10-17
Correct a small type (Tt -> It)

lamosley:
Correct a small type (Tt -> It) which is confusing to non-native English speakers


dougeby:
#sign-off

PR #4657, opened 2024-10-23
Update supported client OS's - Windows Server IoT 2019 for Storage and Windows Server IoT 2022 for Storage  are not supported client OSs

lamosley:
Windows Server IoT 2019 for Storage and Windows Server IoT 2022 for Storage are vendor-specific versions of Windows IoT, and can't be tested. 

I'm adding explicit entries that they are not supported, but if there is a better way to call this out in the documentation, please suggest.


dougeby:
Looks ok. Approving. #sign-off

PR #4660, opened 2024-11-06
Update app-protection-framework-level3.md

sufuruka:
Please confirm which settings are correct to discribe as a Level 3 enterprise high data protection.
If the docs "Action" is wrong, please update the Action to "Wipe data". Or, "Action" is correct, please update the  "(days)" discriprion to (minutes).

"Block access" action is able to set Minutes. It cannot select days. 
"Wipe data" action is able to set Days.


sufuruka:
@Erikre san It seems pending modification docs. I'd appreciate if you couldcheck and proceed this Pull request.

PR #4670, opened 2024-11-12
Update "Retire an Apple ADE device from Intune" section

charlygott:
The layout in Apple Business manager has been changed a bit. Instructions have been updated to the new layout.


charlygott:

PR #4671, opened 2024-11-12

andrewjohnporter:
adding information about monitoring the policy in the Defender portal


Brenduns:
#sign-off 
@andrewjohnporter   Thanks for the great expansion on Monitoring. I added some light formating, etc., and am sending the result through for publication.

PR #4680, opened 2024-11-15
Update overview.md

mrjacobascott:
Updating overview doc to clarify that the Windows of version of the client needs to be an Intune supported version of Windows for comanagement to be available. Based on multiple support cases from customers thinking that tenant attach and comanagment are the same thing, but they are not. Trying to help make it clearer that Windows servers are not in scope for comanagment, even if directly targeted by a comanagement workload.


mrjacobascott:
Hi @gowdhamankarthikeyan, anything needed to push this to live?

dougeby:
The updates look good. Approving. #sign-off

PR #4695, opened 2024-12-15
Update faq.yml

Air-Git:
Delete redundant and incorrect word


frankroj:
Thanks for the contribution @Air-Git, and sorry for the delay in getting to this. I have approved the change, and it should update in the next day or two.

PR #4704, opened 2025-01-19
Update epm-deployment-considerations-ki.md

peteregerton:
Updated file extensions to include .exe as per rules policy section of documentation. https://learn.microsoft.com/en-us/mem/intune/protect/epm-policies#about-windows-elevation-rules-policy


Brenduns:
#sign-off

PR #4707, opened 2025-01-27
Update windows-update-settings.md

brianreidc7:
The Intune portal in English reads "Auto install and restart at a scheduled time" and not "Auto install and restart at scheduled time" (with 'a' missing). Corrected doc to match portal.


Smritib17:
#sign-off

PR #4709, opened 2025-01-30
macOS LOB apps max size limit is 8 GB instead of 2 GB

iye-ms:
macOS LOB apps max size limit is 8 GB instead of 2 GB


Erikre:
#sign-off

Erikre (approved):

Erikre:
#sign-off

PR #4711, opened 2025-01-31
issue: device-profile-assign.md, line 122

Yebbenbe:
This isn't vandalism, I just do not see a way to report inaccurate/missing doc, and i do not know the answer. Line 121 specifically talks about devices  with no associated user and "device groups" - but does not make it clear how to do this. With iOS at least, not having a user associated means the device does not show in Entra, and cannot be added to a group - even if the device shows in InTune and you use the InTune "Groups" component (I know it's just the same React component from Entra, this just doesn't make sense)


MandiOhlinger:
@Yebbenbe Userless devices (enrolled using either Apple Shared iPad or Entra Shared Device Mode) show up in Entra. To target Intune policies to userless devices, you must create a device group like the article outlines.

What are your suggestions to improve this text so it's more clear? Thank you!

MandiOhlinger:
@Yebbenbe I added iOS as an example. If you think any changes should be made, please create another PR. Thank you!

PR #4719, opened 2025-02-19
Update properties-catalog.md

GenerAhl:
Added Windows 11, version 24H2. If we're super prescriptive with supported versions we cannot leave out the latest and greatest.


Smritib17:
The changes are in the docs article. https://learn.microsoft.com/en-us/intune/intune-service/configuration/properties-catalog#supported-platforms

PR #4727, opened 2025-02-27
Update known-issues.md

aiswarya-cloud:
Emailed Juan Carlos Luciani from Microsoft and got this approved to add to public documenation.


frankroj:
Known issues are usually reserved for "bugs" that the product group is actively working on to fix. The submitted content is good, but its just in the wrong place. Possibly move this over to the troubleshooting articles instead?

frankroj:
There are several updates that need to be done regarding properly deleting client records in Autopilot. Work will be performed via the following PR:

https://github.com/MicrosoftDocs/memdocs-pr/pull/17714

PR #4731, opened 2025-03-05
Update managed-software-updates-ios-macos.md

Yebbenbe:
This graph indicates that Managed Software Updates cannot force automatic update with latest version - but isn't this what the "Software Update Settings - Automatic Options" do? It allows specifically setting Automatic updates to Always on. The tooltip basically describes forcing automatic update to latest version... Assuming the graph is deprecated.

What you CAN'T do, from what I can tell, is configure automatic update time periods - so I added that. 

Removed the line saying "More information on these settings is available in the documentation section for the Software Update Settings declarative configuration." because it's not - that link just brings you to a barebones overall Apple Device Config item list, which has nothing but links to Apple documentation. Frankly I get really frustrated with how many different tabs always end up open when I try to read Microsoft documentation, so I just replaced it with a summary of what these do instead of another link for people to open, lose track of, and feel frustrated about. 

Hope these help

### DIFF
--- a/memdocs/intune-service/protect/managed-software-updates-ios-macos.md
+++ b/memdocs/intune-service/protect/managed-software-updates-ios-macos.md
@@ -73,6 +73,10 @@ Use the following information to help you decide which policy type to use.
 | macOS | ✅ | ❌ |
 | &nbsp;|&nbsp; | &nbsp;|
 | **Auto deploy latest update** | &nbsp; | &nbsp; |
+| iOS/iPadOS | 18.0+ | ✅ |
+| macOS | ❌ | ✅ |
+| &nbsp;|&nbsp; | &nbsp;|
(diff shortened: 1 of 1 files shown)


beflamm:
Hi @Yebbenbe - thanks for the suggestions. The automatic actions will not 'force' an update, they only attempt to install overnight/when the device is inactive. The update window is misleading because MDM update policies can't guarantee that an update is installed within the configured window, which is only used to tell Intune that it can send an update within that timeframe. Ack the feedback on the bad link.

@Smritib17 This commit can be closed, I'll include these changes in the next software updates doc release - thanks!

beflamm:
@Yebbenbe managed updates (DDM) allows admins to force install an update on a specific day/time
image

PR #4734, opened 2025-03-07
Update device-compliance-partners.md

kevsully:
CLOMO MDM raised that their branding requires the calpitilization.


dougeby:
#sign-off

PR #4737, opened 2025-03-10
Update device-enrollment-program-enroll-macos.md

AnyaNovicheva1:
paused rollout, need to add that into comms for customers


lenewsad:
#sign-off

PR #4740, opened 2025-03-14
Update whats-new.md

Mads-Melsen:
There was a bullet point with a "-" but no text.


Brenduns:
@Mads-Melsen   Thanks for this PR to clear that extra and misleading bullet point.  There are only the two options.

#sign-off

PR #4745, opened 2025-03-17
Update workloads.md

dereklangMSFT:
Clarification to windows update workload informing customers that they need to disable the software updates client setting from Configuration Manager after moving this workload. This guidance exists in the WUFB documentation, but it needs to be added to our docs because this is a common step that customers do not know about


mestew:
Thanks @dereklangMSFT

PR #4766, opened 2025-04-16
Update microsoft-intune-app-linux.md to add information for Fedora 42

dmac-au:
Fedora 42 no longer supports java-11-openjdk from the standard repositories. Updated instructions to provide information on how to install on Fedora 42 and above.

I'm not sure if this matters to the official documentation as Fedora isn't supported officially, but the below helped me upgrade from Fedora 41 to 42 while keeping the java jdk installed.


dmac-au:
I'm closing this PR as I also had to create a symlink to the correct jdk home and that makes it more difficult for a normal end user. 

Hopefully this doesn't mean that java 11 won't be available in RHEL 10 when it releases.

PR #4773, opened 2025-04-24
Update whats-new.md - typo fix

rhall91:
Correcting Andriod to Android


Brenduns:
#sign-off 

 @rhall91 Thanks for this PR and fix for the Android typo. Having seen that thoes two flew by the normal checks, I ran a search on the entire memdocs docs repository to ensure no other instances snuck through. Those two were the only ones.

PR #4777, opened 2025-05-01
Update create-json-file.md

amaclarencgi:
Write permissions aren't required to produce the autopilot profile JSON files.


frankroj:
I will need to check with PG before making these changes.

PR #4795, opened 2025-06-12
Update apple-account-driven-user-enrollment.md

rishitasarin-ms:
remove known issues section. 

remove this issue for now and if/when we document how to enable ESSO then we can re-add it then. We currently don't document how to enable ESSO because it is not a viable solution for customers. Once apple fixes ESSO, we will add documentation on how to enable it as well as this known issue.


lenewsad:
#sign-off

PR #4824, opened 2025-08-27
Update platform-sso-macos.md to clarify FileVault policy setting

cgryba:
Line 232 needs clarification that this setting only applies when the authentication method is set to "Password".

In Intune, when hovering over the "i" icon beside the "FileVault Policy" setting, you are told: **The policy to apply when using Platform SSO at FileVault unlock on Apple Silicon Macs. Applies when 'AuthenticationMethod' is `Password`.**


MandiOhlinger:
@cgryba Thank you!

PR #5034, opened 2026-04-16
Fixed dead links to macOS LAPS docs

jpjm:
I am not sure that the new link is a total replacement for the previous article, but it's better than a 404. I will let you make your own call on this.


jpjm:
I got the path wrong in my PR, the doc I was trying to link is https://github.com/MicrosoftDocs/memdocs/blob/main/intune/device-security/laps/setup-macos.md

paolomatarazzo:
Already fixed in a different PR

PR #5039, opened 2026-04-17
Clarify device re-registration behavior

iye-ms:


intune/device-configuration/settings-catalog/configure-platform-sso-macos.md
iye-ms, line 1: Here are some of my suggestions:
1. The device will re-register if either Authentication method OR use shared device keys setting is changed. Should not be AND. And the key is to emphasize that for "devices already assigned with PSSO policy", re-registration happens when the settings are changed. But right now it sounds like "all the enrolled devices" will re-register. This is not a big deal but want to bring up and see if any wording can be improved here. 

2. "For the other settings you add **or change**, if the profile is unassigned and reassigned the Platform SSO policy, the device reregisters."

MandiOhlinger: Much more clearer. Thank you!

MandiOhlinger (approved):

PR #5063, opened 2026-04-28
Add eSIM activation server URLs for carriers

CBrasga:
Added eSIM activation server URLs for T-Mobile, Verizon, and AT&T to the update cellular data plan documentation.


paolomatarazzo:
Thank you for the contribution, @CBrasga. Unfortunately, I can't approve this change. We don't want to start maintaining a list of carriers, as it would require too much effort.

PR #5076, opened 2026-05-04
Document BitLocker PIN modification event logging

PraveenGururaj:
Added information about BitLocker pre-boot PIN modification events and their logging in Event Viewer.


Brenduns (approved):
Looks good. Thanks for the tip providing more insight to the PIN changes.

PR #5089, opened 2026-05-11
Correct threat level description in compliance policy

matthewjlevy:
https://learn.microsoft.com/en-us/intune/device-security/mobile-threat-defense/create-compliance-policy

In Intune, the setting is "Clear" and not "Secured"
image


Brenduns:
@matthewjlevy   Thanks for this PR, it looks good. 
#sign-off

Digging in, it appears that Defender changed the label for Secure to Clear some time back, and this reference never recieved the update.  Based on this, I'll also do a sweep of the Intune content to ensure any other references like this we might have are also updated.

PR #5090, opened 2026-05-11
Remove access policies entry

matthewjlevy:
## Documentation issue

The article says that **Access Policies** is a supported resource type for Multi Admin Approval:

https://github.com/MicrosoftDocs/memdocs/blob/main/intune/fundamentals/role-based-access-control/multi-admin-approval.md?plain=1#L30

Specifically, line 30 states:

> Access Policies - Applies to creating or managing multiple administrative approval policies.

## Observed behavior

In my tenant (Service release 2604), when I go to create a Multi Admin Approval access policy, I do **not** see **Access Policies** as an available **Profile type** / resource to protect.
image

## Request

Please clarify one of the following in the documentation:

- whether **Access Policies** is still in rollout / preview / tenant-dependent availability, or
- whether the article is incorrect and this resource type is not generally available, or
- what prerequisites are required for it to appear.

## Suggested doc improvement

If this feature is not universally available, please add a note near the supported resource list indicating that the available resource types can vary by tenant and rollout status.


Brenduns:
@matthewjlevy   Thanks for the PR.  The PM and I reviewed, discussed.  The removal from the bullet point list makes sense, but we're adding a clarification in the line that follows it: 

Changes to access policies require a second administrator's approval. Because this resource type is automatically protected, it isn't available as a selectable profile type when you create an access policy.

PR #5114, opened 2026-05-21
Update deprecation date from far in the future to 2023 (which is more…

stolky:
… likely the right date)version for resource access feature

deprecation in 2203 seems a bit far in the future
this date is likely a typo, and should be 2023
(not that big a deal but a bit sloppy)


Yupiter2005:
Hello @stolky!

Thanks for your contribution!
It's not a typo but a real ConfigMgr version from 03rd month of year 2022, hence 2203. See [ConfigMgr 2203 Checklist](https://learn.microsoft.com/en-us/intune/configmgr/core/servers/manage/checklist-for-installing-update-2203).

PR #5123, opened 2026-05-27
Update sms_supsyncstatus WMI class with possible values

PraveenGururaj:
Added possible values for last synchronization state.


Yupiter2005 (approved):
Thanks for your contribution!

Note that MSFT employees can contribute to private memdocs-pr repo.

PR #5133, opened 2026-06-01
Update ref-settings-android.md

Yebbenbe:
added Important info

### DIFF
--- a/intune/app-management/protection/ref-settings-android.md
+++ b/intune/app-management/protection/ref-settings-android.md
@@ -13,7 +13,7 @@ This article describes the app protection policy settings for Android devices. T
 There are three categories of policy settings: data protection settings, access requirements, and conditional launch. In this article, the term *policy-managed apps* refer to apps that are configured with app protection policies.
 
 > [!IMPORTANT]
-> The Intune Company Portal is required on the device to receive App Protection Policies for Android devices.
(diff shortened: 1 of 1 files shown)


Yebbenbe:
Scenarios that trigger Admin Requirement : 

Require device lock → Low Complexity → Block access

I'll add more as I make sense of them. :/

paolomatarazzo:
@Yebbenbe Thanks for the contribution! I appreciate the additional context you’re trying to provide here.
I’m going to decline this update as written for a couple of reasons:
1. The current [!IMPORTANT] callout is intended to highlight a critical requirement, and the added content introduces subjective language (for example, “may frighten users”) and non-specific statements that don’t align with Microsoft Learn style guidelines.
2. Important callouts should remain concise, factual, and focused. The additional details also mix requirement and explanation in a way that reduces clarity.
3. 
If you feel the behavioral differences on Android are important to call out, I’d recommend capturing that as separate explanatory text, using neutral and precise wording.
For example:

```
> [!IMPORTANT]
> The Intune Company Portal is required on the device to receive app protection policies for Android.
> Installing the Company Portal does not enroll the device.
```

And then, outside the callout:

```
On Android, users may be prompted to grant additional permissions (such as device administrator privileges) when using apps like Outlook. This behavior differs from the experience on iOS.
```

Happy to take another look if you’d like to revise along these lines.

PR #5138, opened 2026-06-03
Update configure-precache-content.md

grafalk:
Are you sure that is not mis in this part ? 

> _For a task sequence that installs an OS image, don't use the deployment option to **Download content locally when needed by the running task sequence**. When the task sequence wipes the disk before it applies the OS image, it removes the client cache. Since the content is gone, the task sequence fails._

In my opinion must be : 

> For a task sequence that installs an OS image, don't use the deployment option to **Download all content locally before starting task sequence**.....

Because in this case client download all content, then disk will be wipe and TS failed.

In short need to replace  "Download content locally when needed by the running task sequence" on "Download all content locally before starting task sequence"

Am I right ? :)

### DIFF
--- a/intune/configmgr/osd/deploy-use/configure-precache-content.md
+++ b/intune/configmgr/osd/deploy-use/configure-precache-content.md
@@ -103,7 +103,7 @@ For example, the following **Upgrade OS** step uses the English version:
 - On the **Distribution Points** tab, configure the **Deployment options** settings. If the content isn't pre-cached before a user starts the installation, the client uses these settings.
 
     > [!IMPORTANT]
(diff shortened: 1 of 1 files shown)


v-dirichards:
@sccmavenger 

Can you review the proposed changes?

**Important**: When the changes are ready for publication, adding a `#sign-off` comment is the best way to signal that the PR is ready for the review team to merge.
 
#label:"aq-pr-triaged"
@MicrosoftDocs/public-repo-pr-review-team

Yupiter2005 (approved):
@grafalk had to read it few times before understanding the change. Thanks for your contribution, approving!

PR #5195, opened 2026-07-29
Update boundary-groups-software-update-points.md

BernEdel:
Screenshot 2026-07-29 135800
There was a typo in the high-level guidance item 4: CGM instead of CMG


v-dirichards:
Looks good, thanks @BernEdel!

